PR #2, opened 2018-05-30
update riscv-coq

andres-erbsen:
for https://github.com/samuelgruetter/riscv-coq/pull/3, also fixes build on coq 8.8


Makefile
```diff
@@ -3,7 +3,7 @@ default_target: all
 
 COQC=$(COQBIN)coqc
 
-EXPECTED_COQC_VERSION := 8.7.2
+EXPECTED_COQC_VERSION ?= 8.7.2
```
samuelgruetter: nice trick

Issue #9, opened 2018-06-15
debugging typeclass resolution in semantics


andres-erbsen:
As a part of an experiment I tried modeling continuations with explicit stacks. In ExprImp.v after defining `stmt`, the following code

~~~~~
  Inductive cont: Set :=
    | CLoad(x: var)(addr: expr): cont
    | CStore(addr val: expr): cont
    | CSet(x: var)(e: expr): cont
    | CIf(cond: expr)(bThen bElse: cont): cont
    | CWhile(cond: expr)(body: cont): cont
    | CSeq(s1 s2: cont): cont
    | CSkip: cont
    | CCall(binds: list var)(f: func)(args: list expr)
    | CStack(st: state)(c: cont)(binds rets: list var).

  Section WithEnv.
    Context {env} {funcMap: Map env func (list var * list var * stmt)} (e:env).
    Fixpoint eval_cont(f: nat)(st: state)(m: mem)(s: cont): option (state * mem) :=
      match f with
      | 0 => None (* out of fuel *)
      | S f => match s with
        | CLoad x a =>
            a <- eval_expr st a;
            v <- read_mem a m;
            Return (put st x v, m)
        | CStore a v =>
            a <- eval_expr st a;
            v <- eval_expr st v;
            m <- write_mem a v m;
            Return (st, m)
        | CSet x e =>
            v <- eval_expr st e;
            Return (put st x v, m)
        | CIf cond bThen bElse =>
            v <- eval_expr st cond;
            eval_cont f st m (if weq v $0 then bElse else bThen)
        | CWhile cond body =>
            v <- eval_expr st cond;
            if weq v $0 then Return (st, m) else
              p <- eval_cont f st m body;
              let '(st, m) := p in
              eval_cont f st m (CWhile cond body)
        | CSeq s1 s2 =>
            p <- eval_cont f st m s1;
            let '(st, m) := p in
            eval_cont f st m s2
        | CSkip => Return (st, m)
        | CCall binds fname args =>
          fimpl <- get e fname;
          let '(params, rets, fbody) := fimpl in
          argvs <- option_all (map (eval_expr st) args);
          st0 <- putmany params argvs empty;
          st1m' <- eval_cont f st0 m fbody;
          let '(st1, m') := st1m' in
          retvs <- option_all (map (get st1) rets);
          st' <- putmany binds retvs st;
          Return (st', m')
        | CStack stf cf binds rets =>
          stf'm' <- eval_cont f stf m cf;
[...]

PR #14, opened 2018-06-21
[do not merge] semantics-refactor

samuelgruetter:
creating a PR, hoping that it will let me comment on several commits at a time...

### DIFF
--- a/src/ExprImp.v
+++ b/src/ExprImp.v
@@ -64,6 +64,17 @@ Section ExprImp1.
     | CSeq(s1:cont) (s2: stmt)
     | CStack(st: state)(c: cont)(binds rets: list var).
   End cont. Arguments cont : clear implicits.
+  Lemma cont_uninhabited (T:Set) (_:T -> False) (X : cont T) : False. Proof. induction X; eauto. Qed.
+  Fixpoint lift_cont {A B : Set} (P : A -> B -> Prop) (a : cont A) (b : cont B) {struct a} :=
+    match a, b with
+    | CSuspended a, CSuspended b =>
+      P a b
+    | CSeq a sa, CSeq b sb =>
+      lift_cont P a b /\ sa = sb
+    | CStack sta a ba ra, CStack stb b bb rb =>
+      (forall vss, agree_on sta vss stb) /\ lift_cont P a b /\ ba = bb /\ ra = rb
+    | _, _ => False
+    end.
 
   Definition ioact : Set := (list var * ioaction * list (word (wXLEN))).
   Definition ioret : Set := (list var * list (word wXLEN)).
--- a/src/ExprImp.v
+++ b/src/ExprImp.v
@@ -22,7 +22,7 @@ Section ExprImp1.
   Notation func := (@name FName).
 
 
-  Context {state: Type}.
+  Context {state: Set}.
   Context {stateMap: Map state var (word wXLEN)}.
   Context {vars: Type}.
   Context {varset: set vars var}.
@@ -43,7 +43,8 @@ Section ExprImp1.
     | SWhile(cond: expr)(body: stmt): stmt
     | SSeq(s1 s2: stmt): stmt
     | SSkip: stmt
-    | SCall(binds: list var)(f: func)(args: list expr).
+    | SCall(binds: list var)(f: func)(args: list expr)
+    | SExtern.
 
   Fixpoint eval_expr(st: state)(e: expr): option (word wXLEN) :=
     match e with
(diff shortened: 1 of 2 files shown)


src/ExprImp.v
```diff
@@ -1,201 +1,453 @@
...
-    | SIf(cond: expr)(bThen bElse: stmt): stmt
-    | SWhile(cond: expr)(body: stmt): stmt
-    | SSeq(s1 s2: stmt): stmt
-    | SSkip: stmt
-    | SCall(binds: list var)(f: func)(args: list expr).
-
-  Fixpoint eval_expr(st: state)(e: expr): option (word wXLEN) :=
+Set Universe Polymorphism.
+Unset Universe Minimization ToSet.
+(* TODO: should we do this?
```
samuelgruetter: what do these two lines do, and why do you need it?

src/ExprImp.v
```diff
@@ -1,201 +1,453 @@
...
+Set Printing Projections.
+
+Require compiler.Common.
+
+Local Notation "'bind_Some' x <- a ; f" :=
+  (match a with
+   | Some x => f
+   | None => None
+   end)
+    (right associativity, at level 60, x pattern).
```
samuelgruetter: Why is this a local notation? Couldn't you put this into a library file?

src/ExprImp.v
```diff
@@ -1,201 +1,453 @@
...
+  (* RecordImport Imp (* COQBUG(https://github.com/coq/coq/issues/7808) *) *)
+  Local Notation expr := Imp.(_expr).
+  Local Notation stmt := Imp.(_stmt).
+  Local Notation cont := Imp.(_cont).
+  Local Notation interp_expr := Imp.(_interp_expr).
+  Local Notation interp_stmt := Imp.(_interp_stmt).
+  Local Notation interp_cont := Imp.(_interp_cont).
+
+  (* TODO: record ImpParametersOK *)
+  Context {mword_eq_dec : DecidableRel (@eq (Imp.varname p))}.
```
samuelgruetter: Why would you separate ImpParameters and ImpParametersOk?
Why don't you use DecidableEq?

src/ExprImp.v
```diff
@@ -204,12 +456,15 @@ Section ExprImp1.
     | SWhile c body => (allVars_expr c) ++ (allVars_stmt body)
     | SSeq s1 s2 => (allVars_stmt s1) ++ (allVars_stmt s2)
     | SSkip => []
-    | SCall binds _ args => binds ++ List.fold_right (@List.app _) nil (List.map allVars_expr args)
+    | SCall binds _ args | SIO binds _ args => binds ++ List.fold_right (@List.app _) nil (List.map allVars_expr args)
     end.
 
+  Context {set_varname: Type}.
+  Context {varset: set set_varname varname}.
```
samuelgruetter: Should these two go into ImpParameters too?

src/ExprImp.v
```diff
@@ -1,201 +1,453 @@
...
+   | None => None
+   end)
+    (right associativity, at level 60, x pattern).
+
+Module Imp. (* TODO: file *)
+
+Record ImpParameters :=
+  {
+    mword : Type;
+    mword_nonzero : mword -> bool;
```
samuelgruetter: abstracting over the word type to hide the dependent type is nice, I meant to do this for a while already

src/ExprImp.v
```diff
@@ -1,201 +1,453 @@
...
 
-  Section WithEnv.
-    Context {env} {funcMap: Map env func (list var * list var * stmt)} (e:env).
-    Fixpoint eval_stmt(f: nat)(st: state)(m: mem)(s: stmt): option (state * mem) :=
+  Section cont.
+    Context {T : Type}.
+    Local Inductive cont : Type :=
+    | CSuspended(_:T)
+    | CSeq(s1:cont) (s2: stmt)
+    | CStack(st: varmap)(c: cont)(binds rets: list varname).
```
samuelgruetter: What's the meaning of these 3 cases?
How much do we commit to by putting this into the semantics?

src/ExprImp.v
```diff
@@ -1,201 +1,453 @@
...
+  End cont. Arguments cont : clear implicits.
+
+  Local Definition ioact : Type := (list varname * actname * list mword).
+  Local Definition ioret : Type := (list varname * list mword).
+  
+  Section WithFunctions.
+    Context (lookupFunction : funname -> option (list varname * list varname * stmt)).
+    Local Fixpoint interp_stmt(f: nat)(st: varmap)(m: mem)(s: stmt): option (varmap*mem*option(cont ioact)) :=
```
samuelgruetter: Is there a way to instantiate ioact with something "empty" such that the existing proofs will still go through?

src/ExprImp.v
```diff
@@ -1,201 +1,453 @@
...
+  Record ImpType :=
+    {
+      _expr : Type;
+      _stmt : Type;
+      _cont : Type -> Type;
+      _interp_expr : forall (st: varmap)(e: expr), option (mword);
+      _interp_stmt : forall (lookupFunction : funname -> option (list varname * list varname * stmt))(f: nat)(st: varmap)(m: mem)(s: stmt),       option (varmap*mem*option(cont ioact));
+      _interp_cont : forall (lookupFunction : funname -> option (list varname * list varname * stmt))(f: nat)(st: varmap)(m: mem)(s: cont ioret), option (varmap*mem*option(cont ioact));
+    }.
+  Definition Imp : ImpType := Build_ImpType expr stmt cont interp_expr interp_stmt interp_cont.
```
samuelgruetter: yet another level of wrapping and indirection :wink: -- why is this needed?

src/ExprImp.v
```diff
@@ -1,201 +1,453 @@
...
+
+    Imp.interp_binop := compiler.Op.eval_binop;
+    Imp.load := Memory.read_mem;
+    Imp.store := Memory.write_mem;
+
+    Imp.varmap := p.(varmap);
+    Imp.varmap_operations := p.(varmap_operations);
+  |}.
+Definition RISCVImp (p:RISCVImpParameters) : Imp.ImpType (ImpParameters_of_RISCVImpParameters p)
+  := Imp.Imp (ImpParameters_of_RISCVImpParameters p).
```
samuelgruetter: By making the right typeclass instances into available, you'd get this for free...

src/ExprImp.v
```diff
@@ -1,201 +1,453 @@
...
+                   RISCVImp.bw
+                   RISCVImp.varname
+                   RISCVImp.funname
+                   RISCVImp.actname
+                   RISCVImp.varmap
+                   RISCVImp.varmap_operations
+         ] in
+    fun bw varname funnname actname varmap varmap_operations =>
+      RISCVImp (Build_RISCVImpParameters bw varname funnname actname varmap varmap_operations).
+*)
```
samuelgruetter: This looks suspicious, I guess you'll need it at some point... Will you have to do this to make sure "rewrite" finds matching terms? And is there a risk that the typechecker isn't happy unless you destruct a parameters record?

src/ExprImp.v
```diff
@@ -218,100 +473,57 @@ Section ExprImp1.
...
     intros.
     induction s; simpl in *.
-    - set_solver.
-    - set_solver.
-    - apply singleton_set_spec in H. auto.
-    - apply union_spec in H.
+    { set_solver. }
+    { set_solver. }
+    { apply singleton_set_spec in H. auto. }
+    { apply union_spec in H.
```
samuelgruetter: Just curious, why do you prefer `{` over bullets?

andres-erbsen: because brackets for one goal do not force the choice of delimiters for other goals. Here I can write 1,2: in the last case.

src/ExprImp.v
andres-erbsen: We won't need it, it was just a test. If record destruction would become an issue we could just turn on primitive projections for that record.

src/ExprImp.v
andres-erbsen: No indirection here, just exporting. In Ocaml this would be in an mli file.

src/ExprImp.v
andres-erbsen: yep.

src/ExprImp.v
andres-erbsen: This is in the refactor PR by accident, let's discuss it separately from the actual refactor.

src/ExprImp.v
andres-erbsen: No, it is not necessary for defining the semantics of the language.

src/ExprImp.v
andres-erbsen: Separete to allow for parallel compilation of proofs (and in other use cases, reification of the code that doesn't have proofs in it).

src/ExprImp.v
andres-erbsen: We could, but I don't want to have the readers of the semantics guessing what `<-` means.

src/ExprImp.v
andres-erbsen: Actually, it is more than exporting. Not making a record of these definitions would mean that the parameters record would be duplicated for each definition in `Imp`. I think https://people.csail.mit.edu/jgross/personal-website/papers/category-coq-experience-itp-submission.pdf explains how that results in sadness and slowness.

src/ExprImp.v
andres-erbsen: Non-universe-polymorphic definitions are depraceted, we use polymorphic ones. `Universe Minimization ToSet` is a backwards compatability mode for that.
`Primitive Projections` means that you never need to destruct records to get something to typecheck.

andres-erbsen:
I am not happy with the boilerplate either. But what else would we do? Using the module system would have less boilerplate but the boilerplate would be more arcane; using typeclasses breaks typechecking errors and this type-driven development in most cases where I have seen it used, and often tactics end up being sensitive to syntactic details of typeclass inference output (I had to change some tactics here because of that).

andres-erbsen:
Full version with desugaring of modules into records at source level: https://github.com/mit-plv/bedrock2/tree/refactor-even-more-boilerplate
Stuck attempt using the old module system: https://github.com/mit-plv/bedrock2/tree/modules-boilerplate

andres-erbsen:
I think the remaining boilerplate can be removed using these two changes:

1. removing `ExprImpInterface` gets rid of some boilerplate at the cost of not being able to pass around all `Imp` definitions as one unit and at the further cost of multiplicative blowup if for each additional parameter next to `ImpParameters` (as opposed to inside of it). Currently we don't do either anywhere.
2. replacing ahead-of-time `RecordImport` (or after (1), manual specialization of individual functions) with  time-of-use typeclass resolution gets rid of the python-generated `RecordImport` boilerplate, at the cost of less predictable typechecking (especially errors, see https://github.com/mit-plv/bedrock2/issues/9).

If both are applied, we get https://github.com/mit-plv/bedrock2/tree/typeclasses-without-functors

PR #15, opened 2018-06-29
[do not merge] starting register allocation and sets/maps refactoring

samuelgruetter:
This diff contains two things
* a sketch of linear scan register allocation, probably with many bugs :wink:
* refactoring of sets and maps to add more functions, and to turn the set/map type from a type class argument into a member of the type class.

@andres-erbsen if there's any aspect in this diff where you want me to take a different path, now is time to speak up :wink:


.travis.yml
```diff
@@ -2,6 +2,10 @@ language: generic
 sudo: required
 dist: trusty
 
+branches:
```
andres-erbsen: does this still run on pull requests?

bedrock2/src/FlatImp.v
```diff
@@ -17,10 +17,10 @@ Section FlatImp1.
   Notation func := (@name FName).
   Existing Instance eq_name_dec.
 
-  Context {state: Type}.
-  Context {stateMap: Map state var (word wXLEN)}.
-  Context {vars: Type}.
-  Context {varset: set vars var}.
+  Context {stateMap: MapFunctions var (word wXLEN)}.
```
andres-erbsen: why `MapFunctions` not `Map`? It contains the type as well, right?

bedrock2/src/FlatImp.v
```diff
@@ -165,10 +167,10 @@ Section FlatImp1.
       eval_stmt (S f) st m1 (SCall binds fname args) = Some p2 ->
       exists params rets fbody argvs st0 st1 m' retvs st',
         get e fname = Some (params, rets, fbody) /\
-        option_all (map (get st) args) = Some argvs /\
-        putmany params argvs empty = Some st0 /\
+        option_all (List.map (get st) args) = Some argvs /\
+        putmany params argvs empty_map = Some st0 /\
```
andres-erbsen: :+1:

.travis.yml
samuelgruetter: the "merge build" still runs (as you can see below), but the "branch build" does not, and I guess that's exactly what we want

bedrock2/src/FlatImp.v
samuelgruetter: Naming is a hard problem! If I want it to call `Map`, then I have to call the corresponding thing for sets `Set`, but that name is already taken :stuck_out_tongue:

andres-erbsen:
Could you merge the maps part so that I can use it in other branches? I'll copy-paste it for now.

PR #16, opened 2018-07-28
[do not merge, but fork] Refactoring

samuelgruetter:
@andres-erbsen this is the renaming we discussed, so you can fork from this branch, and work in the `bedrock2/src` directory, whose only file currently is your `ImpToC.v`.

In the `compiler` directory, everything compiles, but a lot of code is commented out, and that's why I don't want to merge it into master yet, so that if Coq devs use master to benchmark omega vs lia, they'll get code where everything runs, but I hope to merge soon.


samuelgruetter:
It fails on 8.7 because `Coq.Numbers.DecimalZ` and `Coq.Numbers.DecimalString` are missing, and
`(x pattern)` in Notations doesn't work. I don't know an easy fix for this (do you? PRs welcome :wink:), so I decided to only partially support 8.7, where "partially" currently means that the subprojects `bbv` and `riscv-coq` are supported, but the code build on top of it is not.

I think of "allow failure" for 8.7 as a simple way of telling me how good or bad we are at 8.7 support, and if it turns out to be easy to improve something, such as [this fix](https://github.com/mit-plv/bbv/commit/89ea20df6c64e1668faf6bd8052722f9cc19a8a6) which just took me 1 minute and made `bbv` and `riscv-coq` work again, I'll do it, otherwise I don't bother :wink:.

I'll remove the 8.7 compatibility claim from the README if we don't find an easy fix in the next few days.

PR #20, opened 2018-08-21
link to public version of bbv library

maxsnew:
Previous version linked to MIT's github, which requires an MIT account to access.


samuelgruetter:
Thanks!

Issue #19, opened 2018-08-21
bedrock2.Syntax.parameters needs a better name


samuelgruetter:

andres-erbsen:
Why?

samuelgruetter:
It's confusing to have so many types with the same name, and Coq will print the parameters of the last imported module as `parameters` rather than `MyModule.parameters`.

Would renaming them to `MyModule_parameters` work? In particular, would `StringNames` still work, or does it have to override `Syntax.parameters`?

andres-erbsen:
I would instead recommend not importing the modules, but if changing the name really makes it for you, go for it.

PR #21, opened 2018-08-22
separate names for the various parameter typeclasses

samuelgruetter:
Fixes https://github.com/mit-plv/bedrock2/issues/19.

@andres-erbsen is this acceptable for you? :wink: 

My main objection to the current naming is that `StringNames.v` overrides `Syntax.parameters`, whereas the naming in this PR just calls them `Syntax_parameters` or `StringNames_parameters`.

PR #22, opened 2018-08-22
don't change Syntax.parameters when StringNames is imported

samuelgruetter:
More bikeshedding :wink:

I guess this is what I actually wanted.


samuelgruetter:
Fixes https://github.com/mit-plv/bedrock2/pull/19, supersedes https://github.com/mit-plv/bedrock2/pull/21.

andres-erbsen:
Sure, I don't object, but I also don't understand how this is better than just not importing StringNames.

samuelgruetter:
Indeed, if you never import StringNames, I don't have a problem. But when reading StringNames (and currently being in a mood where I worry about module features I don't understand), I don't immediately see whether StringNames is never imported, and might get afraid that this solution *relies* on module merging, which would be a bit scary...

PR #24, opened 2018-08-23
Import prim token notations before using them

### LINKED ISSUE #23: bedrock2 is incompatible with numeral notations
I am preparing a PR to fix the code in bedrock2 itself, but please merge https://github.com/mit-plv/bbv/pull/20 and https://github.com/mit-plv/riscv-coq/pull/2 asap and update the submodules in bedrock2.

JasonGross:
This is required for compatibility with
https://github.com/coq/coq/pull/8064, where prim token notations no longer
follow `Require`, but instead follow `Import`.

c.f. https://github.com/coq/coq/pull/8064#issuecomment-415493362

Closes #23 

I'm pretty sure this builds, but you should probably wait for the CI to be sure.

### DIFF
--- a/bedrock2/src/BasicC64Semantics.v
+++ b/bedrock2/src/BasicC64Semantics.v
@@ -1,3 +1,4 @@
+Require Import Coq.NArith.BinNatDef.
 Require Import bedrock2.BasicC64Syntax bedrock2.Semantics.
 Require bedrock2.String bedrock2.Map.UnorderedList.
 Require bbv.Word.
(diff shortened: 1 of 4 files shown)


samuelgruetter:
Btw it seems you have an emacs setting which fixes whitespacing on each save. I want this too! How do I enable this?

samuelgruetter:
I was not so thrilled by this solution, but Thomas pointed me to the `whitespace-cleanup` command, so all I had to do was to add this one line to my `.emacs`:

```
(add-hook 'before-save-hook 'whitespace-cleanup)
```

Thanks anyways @JasonGross :wink:

JasonGross:
Note that `(add-hook 'before-save-hook 'whitespace-cleanup)` changes spaces to tabs in some cases, even for lines that are not touched by the current edit, and so does not work for my purposes.

samuelgruetter:
I see, so apparently that's what `ws-butler` solves

Issue #27, opened 2018-09-04
Define a semantics in the bedrock2 directory


gmalecha:
Right now, the `bedrock2` directory contains the syntax, but the `Semantics.v` file doesn't contain an actual semantics (just some auxiliary definitions). As a first guess, I would assume that `ExprImp` needs to be moved to the `bedrock2` directory, though it seems to have a lot of dependencies, so it should probably be split up.

@samuelgruetter Thoughts on this?

samuelgruetter:
This is on my TODO list, but a PR for this would be welcome too :wink:

andres-erbsen:
How about the following hybrid of traditional big-step semantics, Cito-style semantics, and axiomatic weakest-precondition semantics?

~~~~~
Inductive exec
  : cmd (* program *)
  -> trace -> mem -> locals (* initial state *)
  -> (trace -> mem -> locals -> Prop) (* postcondition *)
  -> Prop :=
| skip t m l post (_:post t m l)
  : exec cmd.skip t m l post
| set x e t m l post
      v (_ : eval l e v)
      (_ : post t m (map.put l x v))
  : exec (cmd.set x e) t m l post
| seq c1 c2 t m l post
      mid (_ : exec c1 t m l mid)
      (_ : forall t m l, exec c2 t m l post)
  : exec (cmd.seq c1 c2) t m l post
| while e c t m l post
        v (_ : eval l e v)
        (_ : Logic.or
               (word_test v = false /\ post t m l)
               (word_test v = true /\ exec (cmd.seq c (cmd.while e c)) t m l post))
  : exec (cmd.while e c) t m l post
| interact bindvars action arges t m l post
           args (_ : List.Forall2 (eval l) arges args)
           (output := (m, action, args))
           (_ : forall m rets (t := cons (output, (m, rets)) t),
               guarantee t /\
               rely t -> (bind_ex_Some l <- map.putmany bindvars rets l; post t m l))
  : exec (cmd.interact bindvars action arges) t m l post.
~~~~~

I am guessing reactive and divergent programs could be modeled by replacing `Inductive` with `CoInductive`, but I don't claim any expertise over that.

PR #30, opened 2018-09-10
Proposal: Globals (pre-cursor to Compilation Units)

gmalecha:
This generalizes functions into globals. It still doesn't support function pointers though doing so is relatively simple. In summary,

```coq
funname -> globname

expr := ..
  | global (_ : globname)
```

Because it was simple, I also changed `call` from taking a `funname` to taking an `expr` (so `f` becomes `global f`). Generalizing the semantics completely isn't too difficult. I can add that to this PR if it would be useful. #27 is preventing a complete operational semantics, but if we solve that, it shouldn't be difficult to do this.

### DIFF
--- a/bedrock2/src/WeakestPrecondition.v
+++ b/bedrock2/src/WeakestPrecondition.v
@@ -5,6 +5,7 @@ Require Import Coq.ZArith.BinIntDef.
 Section WeakestPrecondition.
   Context {p : unique! Semantics.parameters}.
   Context (rely guarantee : trace -> Prop) (progress : trace -> trace -> Prop).
+  Variable resolver : globname -> option word.
 
   Definition literal v post : Prop :=
     bind_ex_Some v <- word_of_Z v; post v.
@@ -23,6 +24,8 @@ Section WeakestPrecondition.
         literal v post
       | expr.var x =>
         get l x post
+      | expr.global g =>
+        bind_ex_Some v <- resolver g ; post v
       | expr.op op e1 e2 =>
         expr e1 (fun v1 =>
         expr e2 (fun v2 =>
@@ -46,7 +49,7 @@ Section WeakestPrecondition.
   End WithF.
 
   Section WithFunctions.
-    Context (call : funname -> trace -> mem -> list word -> (trace -> mem -> list word -> Prop) -> Prop).
+    Context (call : globname -> mem -> list word -> (trace -> mem -> list word -> Prop) -> Prop).
     Fixpoint cmd (c : cmd) (t : trace) (m : mem) (l : locals)
              (post : trace -> mem -> locals -> Prop) {struct c} : Prop :=
       match c with
@@ -67,7 +70,7 @@ Section WeakestPrecondition.
       | cmd.seq c1 c2 =>
         cmd c1 t m l (fun t m l => cmd c2 t m l post)
       | cmd.while e c =>
-        exists measure (lt:measure->measure->Prop) (inv:measure->trace->mem->locals->Prop), 
+        exists measure (lt:measure->measure->Prop) (inv:measure->trace->mem->locals->Prop),
         Coq.Init.Wf.well_founded lt /\
         (exists v, inv v t m l) /\
         (forall v t m l, inv v t m l ->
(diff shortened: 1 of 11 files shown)


gmalecha:
Just a note, on top of this definition we would have something like:

```coq
Record module :=
{ definitions : map globname (function + data)
; imports : list globname
; exports : list globname
}.
```

bedrock2/src/WeakestPrecondition.v
```diff
@@ -46,7 +49,7 @@ Section WeakestPrecondition.
   End WithF.
 
   Section WithFunctions.
-    Context (call : funname -> trace -> mem -> list word -> (trace -> mem -> list word -> Prop) -> Prop).
+    Context (call : word -> trace -> mem -> list word -> (trace -> mem -> list word -> Prop) -> Prop).
```
samuelgruetter: `funname` should remain abstract in this file

gmalecha: It is abstract, in this version of the semantics you don't call a `funname`, you call the word that it resolves to.

gmalecha: Further clarification. The `resolve` function maps `globname` to `option word` and `global g` denotes to `resolve g` (if it is defined). So previously, if you wanted to say that "`"swap"` has this specification", you would now say "`resolve "swap"` has this specification".

gmalecha: It wouldn't be difficult to remove this functionality (just don't allow calling expressions), if that is truly desireable. This seems like an easy way to support function pointers within the "be concrete" mantra of bedrock2.

samuelgruetter:
I'd like to see what `compiler.ExprImp.eval_cmd` looks like with this (I think this could be done before or after moving the semantics in `ExprImp.v` from the `compiler` folder to the `bedrock2` folder, whatever seems more convenient).

I still believe that `funname` and `globname` should be two different abstract types (but of course you can have files inside which the two are defined to be the same, if all users of such files are ok with it). But since I'm not working with functions myself at the moment, I'm fine with using the same name for `funname` and `globname` for the moment, and once I start working with functions myself, I'll see whether/why I want to separate them, so be prepared that I'll want to change it, but for the moment I'm ok to stick to the YAGNI principle :wink:

"Because it was simple" is not a good reason for replacing `funname` by `expr` in `call`. Could you make the change in such a way that there's a simple way to restrict the input language back to what we have now for users who wish to do so?

bedrock2/src/CompilationUnit.v
```diff
@@ -0,0 +1,107 @@
+Require Import Coq.Strings.String.
+Require Import Coq.ZArith.BinIntDef.
+Require Import ExtLib.Data.HList.
+Require Import ExtLib.Data.Fin.
+Require Import ExtLib.Data.Map.FMapAList.
```
samuelgruetter: looks like these `ExtLib` dependencies are not needed (but make the build fail)

gmalecha:
@samuelgruetter I've ported `compiler.ExprImp.eval_cmd`. As the diff shows, it is mostly just alpha renaming and adding some cases for the `global` constructor.

If you can think of any reason within the compiler to separate the two, I'd be really interested in knowing what it is.

I've reverted the ability to call expressions per your request. I guess we can also drop the `call_rec` function that I implemented.

samuelgruetter:
In the [latest build](https://travis-ci.com/mit-plv/bedrock2/jobs/144881353), `ExprImp.v` is never compiled, because the build fails before that. Before I spend more time on trying to understand what you're doing, would you mind making sure that `ExprImp.v` is compiled by travis?

gmalecha:
@samuelgruetter The only thing that didn't compile in the bedrock2 directory is the swap example. I've updated that with the latest push. Now what breaks is in the compiler directory.

bedrock2/src/WeakestPrecondition.v
samuelgruetter: I'd like to be able to use the file `WeakestPrecondition.v` with `funname` being instantitated to `string`, while others might want to instantiate it to `word` or to `Z` or whatever. Can you adapt it so that this becomes possible again?

compiler/src/ExprImp.v
```diff
@@ -20,24 +20,27 @@ Section ExprImp1.
...
   Notation var := (@varname (@syntax p)).
-  Notation func := (@funname (@syntax p)).
+  Notation glob := (@globname (@syntax p)).
 
   Context {stateMap: MapFunctions var (mword)}.
   Notation state := (map var mword).
   Context {varset: SetFunctions var}.
   Notation vars := (set var).
 
+  Variable resolve : glob -> option word.
```
samuelgruetter: nitpick: I'd prefer `mword` over `word` here, even though both refer to the same in this context, it's just that this file consistently uses `mword`. Or if you prefer, you could make sure nothing of bbv is imported to avoid confusion with bbv's word, and then replace all occurrences of `mword` by `word`.

gmalecha: That's fine. Does nitpicking mean that you're considering merging this?

PR #34, opened 2018-11-22
tweak locals notation

samuelgruetter:
it's a matter of taste I guess :wink:


andres-erbsen:
Ok. I am thinking that I might change this further to get the locals to be universally quantified instead of being existentially quantified on the left of an arrow. (like currying).

PR #36, opened 2018-11-22
implement cmd.unset for declaring a variable dead

andres-erbsen:
we would prove a lemma that if a program is good with unset, it is also good without unset. Then we would write an unverified thing to insert unset at locations where a variable becomes dead. each program logic proof would start by replacing the original program with the annotated program using the lemma.

### DIFF
--- a/bedrock2/src/Examples/ipow.v
+++ b/bedrock2/src/Examples/ipow.v
@@ -50,9 +50,11 @@ Local Coercion var (x : string) : expr := expr.var x.
 Definition ipow := ("ipow", (["x";"e"], (["ret"]:list varname), bedrock_func_body:(
   "ret" = 1;;
   while ("e") {{
-    if ("e" .& 1) {{ "ret" = "ret" * "x" }};;
+    "t" = "e" .& 1;;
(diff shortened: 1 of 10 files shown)


bedrock2/src/Examples/ipow.v
```diff
@@ -50,9 +50,11 @@ Local Coercion var (x : string) : expr := expr.var x.
 Definition ipow := ("ipow", (["x";"e"], (["ret"]:list varname), bedrock_func_body:(
   "ret" = 1;;
   while ("e") {{
-    if ("e" .& 1) {{ "ret" = "ret" * "x" }};;
+    "t" = "e" .& 1;;
+    if ("t") {{ "ret" = "ret" * "x" }};;
     "e" = "e" >> 1;;
-    "x" = "x" * "x"
+    "x" = "x" * "x";;
+    cmd.unset "t"
```
andres-erbsen: this use of unset is actually necessary, removing it makes the locals map at the beginning of the loop have different variables than at the end of the loop. alternatively, t could be initialized before the loop (e.g. if it is used after the loop as well).

PR #44, opened 2018-11-26
Optimization: Boolean expression compiler

stellamplau:
This adds a small optimization when compiling Boolean expressions. 

ExprImp cond/while branching on conditions such as `x < y` are now compiled directly to FlatImp SIf/SLoop expressions, retaining this branching information. This allows direct compilation to RISC-V blt/bge instructions. 

Previously, compiling `if (x < y) ...` would result in something like `t1 := x < y; beqz t1 ...` in RISC-V. Now we can directly branch, i.e. `bge x y ...`. 

All minor (and not so minor) comments are welcome and appreciated!

### DIFF
--- a/compiler/src/FlatToRiscv.v
+++ b/compiler/src/FlatToRiscv.v
@@ -1759,7 +1759,7 @@ Section FlatToRiscv.
         inversion H; clear H
     end.
 
-  Ltac tac1 :=
+  Ltac simplify_bcond :=
     (match goal with
     | |- context[@Bind _ _ _ _ _ _ ] =>
         first [rewrite Bind_getRegister0 |
@@ -1912,7 +1912,7 @@ Section FlatToRiscv.
       (* branch if cond = false (will not branch *)
       eapply runsToStep; simpl in *; subst *.
       fetch_inst. eapply sequence_lemma.
-      destruct cond; repeat tac1.
+      destruct cond; repeat simplify_bcond.
       apply execState_step.
       (*run1step.*)
       (* use IH for then-branch *)
@@ -1929,7 +1929,7 @@ Section FlatToRiscv.
       eapply runsToStep; simpl in *; subst *.
       fetch_inst.
       eapply sequence_lemma.
-      destruct cond; repeat tac1. 
+      destruct cond; repeat simplify_bcond. 
       apply execState_step.
       (*run1step.*)
       (* use IH for else-branch *)
@@ -1944,7 +1944,7 @@ Section FlatToRiscv.
       eapply runsToStep; simpl in *; subst *.
       fetch_inst.
       eapply sequence_lemma.
-      destruct cond; repeat tac1.
+      destruct cond; repeat simplify_bcond.
       apply execState_step.
       run1done.
 
@@ -1958,7 +1958,7 @@ Section FlatToRiscv.
       (* 2nd application of IH: part 2 of loop body *)*)
       eapply runsToStep; simpl in *; subst *.
       fetch_inst. eapply sequence_lemma.
-      destruct cond; repeat tac1.
+      destruct cond; repeat simplify_bcond.
       apply execState_step.
 
       (*run1step.*)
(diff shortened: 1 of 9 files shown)


compiler/src/FlattenExpr.v
```diff
@@ -429,7 +528,105 @@ Section FlattenExpr.
            | E: reg_eqb _ _ = false |- _ => apply reg_eqb_false in E
            end.
 
-  Lemma flattenStmt_correct_aux:
+  Ltac cleanup_options :=
+    repeat (match goal with
```
samuelgruetter: nitpick: `repeat match goal` does not need parentheses

compiler/src/FlattenExpr.v
```diff
@@ -429,7 +528,105 @@ Section FlattenExpr.
...
+    | H : None = Some _ |- _ =>
+        inversion H
+    | H : Some _ = None |- _ =>
+        inversion H
```
samuelgruetter: you can leave this as is, but fyi `discriminate H` would work here too, and the above 4 lines could even be replaced by just `try discriminate`, which finds the contradictory equality on its own

compiler/src/FlattenExpr.v
```diff
@@ -429,7 +528,105 @@ Section FlattenExpr.
...
+    1,2,3:
+      repeat destruct_one_match_hyp; repeat destruct_pair_eqs; subst;
+      pose proof flattenExpr_correct_aux as P;
+      specialize P with (initialM := initialM) (1 := E) (4 := Ev);
+      specializes P; [eassumption | eassumption | ];
+      destruct P as [fuelS0 [initial2L [Evcond G]]];
+      exists fuelS0 initial2L;
+      split; [eassumption| unfold FlatImp.eval_bcond; rewrite G; eauto].
+
+    do 4 destruct_one_match_of_hyp F; repeat destruct_pair_eqs; subst.
```
samuelgruetter: This line opens 3 subgoals, but is not followed by `{` nor bullets, but the second and third subgoal of `specializes P.` further down is "sold" as fourth and fifth subgoal of `specializes P.`
For someone stepping through this code, this is super confusing.

samuelgruetter: As soon as we're on 8.9, I'll start using `Default Goal Selector "!"` to only allow one active goal at a time, so that such a sin would result in a build error :wink:

compiler/src/FlattenExpr.v
```diff
@@ -429,7 +528,105 @@ Section FlattenExpr.
...
+    pose proof flattenExpr_correct_aux as P.
+    specialize P with (env := env) (initialM := initialM) (1 := E) (4 := E1).
+    specializes P; [eassumption | eassumption | ].
+    destruct P as [fuelS0 [initial2L [Evcond G]]].
+
+    pose proof flattenExpr_correct_aux as P.
+    specialize P with (initialL := initial2L) (env := env)
+                      (initialM := initialM) (1 := E0) (4 := E2).
+    pose_flatten_var_ineqs.
+    specializes P.
```
samuelgruetter: Eventually I want to get rid of `lib.LibTacticsMin.`, so please don't use `specializes`. This line should be
```
    edestruct P as [fuelS1 [initial3L [Evcond2 G2]]]; [state_calc..|].
```

compiler/src/FlatToRiscv.v
```diff
@@ -1817,18 +1926,26 @@ Section FlatToRiscv.
 
     - (* SIf/Else *)
       (* branch if cond = 0 (will  branch) *)
-      run1step.
+      eapply runsToStep; simpl in *; subst *.
```
samuelgruetter: This line opens two subgoals, but isn't followed by `{` or bullets. That makes proof scripts much harder to read and maintain. Please ensure that you always only have one focused goal at a time.

compiler/src/FlattenExpr.v
stellamplau: Ah thanks for bringing this to my attention! I'll definitely be more cognizant of this in the future.

stellamplau:
Thanks for the pointers. I've made relevant updates. 

I feel like there should be a cleaner way to prove `flattenBooleanExpr_correct_aux`, and also a better way to avoid duplication in the `flattenExprAsBoolExpr` version of lemmas such as `flattenExprAsBoolExpr_freshVarUsage`.

samuelgruetter:
Thanks @stellamplau, great work!

PR #50, opened 2019-01-23
sketch squarer program with IO

samuelgruetter:
Here's how, according to my current understanding, a squarer program with IO could be specified by abstracting over the kind of IO, and how the IO kind could be instantiated to syscalls or bare-metal MMIO to an ethernet card connected via SPI (inspired by the ethernet driver code you sent me a while ago).
@andres-erbsen does that make sense to you?

### DIFF
--- a/bedrock2/src/Examples/Trace.v
+++ b/bedrock2/src/Examples/Trace.v
@@ -1,0 +1,124 @@
+Require Import Coq.ZArith.ZArith. Open Scope Z_scope.
+Require Import Coq.Lists.List. Import ListNotations.
+Require Import coqutil.Word.Interface.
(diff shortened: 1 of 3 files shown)


andres-erbsen:
this looks reasonable. I think you can use Kleene for repetition if you define the single-step again and go cases separately first. syscalls changing memory should somehow use speration logic, maybe existentially quantify over frame and two versions of the changed part and two versions of memory, tie both sides using sep... sounds verbose but I think that's what we mean.

samuelgruetter:
Manually merged, and added a new commit on master: https://github.com/mit-plv/bedrock2/commit/95c91b3c4d12ca48379bd7ec8570d72dada77744

PR #57, opened 2019-02-14
Fix build on Windows

JasonGross:
Coq on Windows generally does not support paths like `/cygdrive/d/Documents/GitHub/bedrock2/deps/coqutil/src/Datatypes/HList.v`.  These are the sorts of paths generated by `pwd` / `find` under cygwin, unless you run the initial path through `cygpath -m` (`-m` for mixed, which makes it use forward slashes).  We pipe stderr to `/dev/null` and fallback to `pwd` on platforms where `cygpath` is not in path (i.e., non-windows platforms).

Note that https://github.com/mit-plv/coqutil/pull/2 and https://github.com/mit-plv/riscv-coq/pull/6 will also need to be merged and the submodules will need to be updated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ update_all:
 	git submodule update --recursive --remote
 
 REL_PATH_OF_THIS_MAKEFILE:=$(lastword $(MAKEFILE_LIST))
+ABS_ROOT_DIR:=$(abspath $(dir $(REL_PATH_OF_THIS_MAKEFILE)))
 # use cygpath -m because Coq on Windows cannot handle cygwin paths
 ABS_ROOT_DIR:=$(shell cygpath -m '$(ABS_ROOT_DIR)' 2>/dev/null || echo '$(ABS_ROOT_DIR)')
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -9,7 +9,8 @@ update_all:
 	git submodule update --recursive --remote
 
 REL_PATH_OF_THIS_MAKEFILE:=$(lastword $(MAKEFILE_LIST))
-ABS_ROOT_DIR:=$(abspath $(dir $(REL_PATH_OF_THIS_MAKEFILE)))
+# use cygpath -m because Coq on Windows cannot handle cygwin paths
+ABS_ROOT_DIR:=$(shell cygpath -m '$(ABS_ROOT_DIR)' 2>/dev/null || echo '$(ABS_ROOT_DIR)')
```
samuelgruetter: This looks like a cyclic definition, is it what you meant? Does it actually become an absolute path, or is it just taking the empty string, which happens to denote the current directory, relatively?

JasonGross: I believe it is cyclic if you use delayed assignment (`=`), and sensible if you use eager assignment (`:=`).  I may have swapped which assignment symbol is which kind of assignment, and perhaps this deserves a comment?  Or we could just not mutate the value and use a different variable name the second time.

JasonGross: Oh, oops, I did not mean to delete the line above this one

JasonGross: Oh, oops, I did not mean to delete the line above this one

PR #61, opened 2019-02-23
consolidate bedrock2.Syntax.Basic; port examples

andres-erbsen:
Threading syntax parameters (and coercions, and instances) everywhere is annoying. AFAIK we have no strong arguments in favor of keeping the syntax parametric. The compiler can translate to register names internally, inline assembly can be supported by having a separate environment of named assembly snippets. Removing this flexibility simplifies the infrastructure considerably, and I am reasonably confident that even if we do want to later add inline asm as `actname`, appropriate notations can be used to parse existing code.

This PR is step one of three, it ports all examples to use the same syntax parameters (all variants strings). Next, the compiler would need to be ported. After that, these parameters can be hardcoded in `bedrock2.Syntax`. I think porting the compiler may need to introduce an admit in the function call case.

The removals in the compiler directory is stuff that looked like cruft to me (and I didn't know how to port it), please correct me if I am wrong.

The changes in variable names in proofs are because 0b674d79885d6dff61671cb09e45770204a8ca30 now works for the examples in Demos, reusing names in the bedrock2 program instead of generating fresh names.


andres-erbsen:
@samuelgruetter what do you think of this direction?

PR #70, opened 2019-04-23
simplify expr metrics WIP

andres-erbsen:
@Carotti this is what I had in mind with my last comment in the metrics PR. Does this make send and seem acceptable to you?

(this is not done yet)


andres-erbsen:
no longer feasible given additions to expr language

PR #74, opened 2019-05-13
Fibonacci Server Demo

Carotti:
Includes a slightly stronger Pipeline proof statement for reasoning about locals, memory, program counters and metrics as well as the trace.

The fib_correct proof proves correctness and instruction bounds on the high level program. fib_compiled uses this updated pipeline proof to show that the correctness and instruction bounds also hold for execution on a RISC-V machine.


samuelgruetter:
It seems that in Coq 8.9, [700MB of RAM are sufficient](https://travis-ci.com/mit-plv/bedrock2/jobs/199837718#L2325) for `FibCompiled`, while in Coq master, [more than 6.4GB of RAM would be needed](https://travis-ci.com/mit-plv/bedrock2/jobs/199837717#L2329-L2330). We'll have to investigate this before merging.

Carotti:
Think I figured it out - blia seemed to be a bit more eager on coq master and it ended up trying to directly compute `fib 47` in order to solve a goal where `fib 47` was already in a hypothesis. I've fixed this so it should be a bit more robust. Let's see if that build works a little bit better!

samuelgruetter:
Cool you figured it out! 
Lia is currently being refactored on coq master, and the fact that it tries to simplify `fib 47` seems really bad to me, we should report that. Can you make a minimal example which succeeds on 8.9 and fails on master?

PR #75, opened 2019-05-13
KamiStep_sound should cover all low-level steps

andres-erbsen:
The top-level theorem should prove that we can relate every kami trace to a higher-level trace, instead of being applicable only to the kami traces that have a high-level counterpart. I think the statement before this change was just too weak -- e.g. iiuc it would allow calling the external method "launchMissiles" regardless of what trace property we proved about the program.

I also removed some dependent type hackery from KamiStepR.

Unfortunately I did not manage to fix up the KamiStep sound proof. This is because

(1) the goal, as now stated, is probably false because the processor as modelled there treats program initialization as external effects, whereas it should be internal
(2) I am not familiar with that part of the project
(3) the current proof has a lot of autogenerated hypothesis names in it, and the tactics seem to create hypotheses without care for names

Let me know if the changes I am proposing make sense for you. I am probably not going to spend any more time trying to fix this right now, but I am hoping this draft is helpful.

### DIFF
--- a/processor/src/KamiRiscv.v
+++ b/processor/src/KamiRiscv.v
@@ -174,35 +174,19 @@ Section Equiv.
     subset (riscvTraces init) (prefixes (post_to_traces post)).
   Admitted.
 
-  Definition KamiLabelR (klbl: Kami.Semantics.LabelT) (ev: list Event): Prop.
-  Proof.
-    refine (match FMap.M.find "mmioExec"%string klbl.(calls) with
-            | Some sv => _
-            | None => ev = nil
-            end).
-    destruct sv as [[argT retT] [argV retV]].
-    destruct (decKind argT (Struct (RqFromProc (Z.to_nat width) rv32DataBytes)));
-      [subst|exact False].
-    destruct (decKind retT (Struct (RsToProc rv32DataBytes)));
-      [subst|exact False].
+  Definition KamiSilent (klbl: Kami.Semantics.LabelT) : Prop :=
+    klbl.(calls) = FMap.M.empty _.
+  Definition KamiLabelR (klbl: Kami.Semantics.LabelT) (e : Event) : Prop :=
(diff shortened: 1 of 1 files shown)


processor/src/KamiRiscv.v
```diff
@@ -237,38 +221,38 @@ Section Equiv.
...
+    forall (m1 m2: KamiMachine) (m1': RiscvMachine) (t0: list Event)
+           (post: RiscvMachine -> Prop) klbl,
+      kamiStep m1 m2 klbl ->
       states_related (m1, t0) m1' ->
       mcomp_sat_unit (run1 iset) m1' post ->
       (* Either Kami doesn't proceed or riscv-coq simulates. *)
-      ((m1 = m2 /\ t = nil) \/
-       exists m2', states_related (m2, t ++ t0) m2' /\ post m2').
+      (KamiSilent klbl /\ m1 = m2) \/
+       (exists t, KamiLabelR klbl t /\ exists m2', states_related (m2, cons t t0) m2' /\ post m2').
```
samuelgruetter: Doesn't this exclude the case where Kami makes a step from `m1` to a non-equal `m2` without producing any event?

andres-erbsen: it looks like that, yeah. I'm not sure what it is all about, I just tried to do equivalence transformations.

joonwonc:
I think I applied all the stuffs mentioned in this pull request in this commit: https://github.com/mit-plv/bedrock2/commit/2e50389c79645288f1b78e0dc0b5a47b49b9d1ad 
I redefined some definitions (`KamiLabelR`, etc.) and am quite sure the changes reflect what we've discussed so far.
Please let me know if you find anything strange.

samuelgruetter:
Cool, https://github.com/mit-plv/bedrock2/commit/2e50389c79645288f1b78e0dc0b5a47b49b9d1ad looks good to me :+1:

PR #77, opened 2019-06-01
prove io behavior of main loop

andres-erbsen-sifive:


andres-erbsen-sifive:
on top of https://github.com/mit-plv/coqutil/pull/3

bedrock2/src/Examples/SPI.v
```diff
@@ -63,21 +63,138 @@ Require Import bedrock2.ProgramLogic.
...
+Import coqutil.Tactics.letexists.
+Import TailRecursion.
+Lemma spi_write_ok : Semantics.parameters_ok parameters -> program_logic_goal_for_function! spi_write.
 Proof.
+  intros p_ok.
+  repeat straightline.
+  rename H into Hb.
+  show_program.
+
+  (* WHY do theese parentheses matter? *)
```
andres-erbsen: @JasonGross have you seen this before?

JasonGross: Very strange.  First guess: scoping issue?  What happens if you do `let foo := uconstr:(...) in refine foo`?  Do you need any extra parentheses?  And is `atleastonce` a notation or a definition?

andres-erbsen-sifive: `atleastonce` is an opaque lemma (with a comp[licated depndent type, it is added in this PR). 
```coq
  Fail let x := open_constr: ((atleastonce ["b"] (fun v T M B => True)) _ _ _ _ _ _ _) in idtac.
  (* The term "[];+"b"" has type "list string" while it is expected to have type "list varname". *)
  Fail let x := open_constr: ( atleastonce ["b"] (fun v T M B => True)  _ _ _ _ _ _ _) in idtac.
  (* The term "[];+"b"" has type "list string" while it is expected to have type "list varname". *)
  assert_succeeds (refine ((atleastonce ["b"] (fun v T M B => True)) _ _ _ _ _ _ _)).
```

bedrock2/src/Examples/LAN9250.v
```diff
@@ -184,15 +184,36 @@ Proof.
...
+
+  all : try rewrite word.unsigned_slu by exact eq_refl.
+  all : match goal with |- context[word.unsigned (width:=?w) ?x] => is_var x; replace (word.unsigned (width:=w) x) with (word.wrap (width:=w) (word.unsigned (width:=w) x)) by eapply Properties.word.wrap_unsigned; set (word.unsigned (width:=w) x) as X; clearbody X end.
+  all : try erewrite ?word.unsigned_of_Z.
+  all : cbv [word.wrap].
+  all : change Semantics.width with 32.
+  all : rewrite <-!Z.land_ones by Lia.lia.
+
+
+  (*
```
andres-erbsen-sifive: @samuelgruetter did we ever figure out a "maximal supportable" z bitwise prover?

samuelgruetter: We have `Z.bitblast` in coqutil, which is quite simple and has very few dependencies: https://github.com/mit-plv/coqutil/blob/master/src/Z/bitblast.v
And we have `prove_Zeq_bitwise` which is a bit more complicated, but faster and solves more goals: https://github.com/mit-plv/riscv-coq/blob/master/src/Utility/prove_Zeq_bitwise.v

andres-erbsen-sifive:
@samuelgruetter can you rerun the travis CI? neither of my accounts seems to be able to, so I am just pushing an empty force to do it...

samuelgruetter:
You have to "sign in with GitHub" on travis. Took me some googling to find out the first time I wanted to rerun a build ;)

PR #84, opened 2019-06-28
Bringup1

andres-erbsen-sifive:
lan9250_readword does something on simulation but I haven't inspected the trace yet


andres-erbsen-sifive:
...and the trace looks as expected. Monday morning gonna do this on real hardware.

PR #101, opened 2019-09-24
Use COQPATH for bedrock2 dependencies

JasonGross:
Also allow passing `EXTERNAL_DEPENDENCIES=1` to `make` to make use of a
system-wide installed copy.

Note that Kami used to be bound with `-R` for end2end, but is now bound with the equivalent of `-Q`.  Hopefully this doesn't break anything (I'm building it on my machine to check).


samuelgruetter:
If I understand correctly, there are three desirable properties which would be lost by merging this PR:

1. If an illegal dependency is introduced in a file, `make` is guaranteed to fail.
2. If an illegal dependency is introduced in a file, processing the file in Proof General will fail.
3. The dependency location options passed to Coq by `make` are the same as the dependency location options passed to Coq by Proof General, *by construction*, i.e. in such a way that the two can't get out of sync.

By "illegal dependency", I mean a dependency which violates the intended subproject dependency graph:

image

For example, a file in the `bedrock2` subdirectory depending on a file in `deps/riscv-coq` would be an illegal dependency.

Currently, 1) and 2) are achieved by only passing the `-Q` options which are really needed, and 3) is achieved by auto-generating the `_CoqProject` file from the makefile and passing it both to Proof General and `coq_makefile`.

@JasonGross can you restore these three desirable properties?

JasonGross:
I have restored (1) and (2) by splitting up the `.dir-locals.el` file into a separate one for each subproject, as well as setting `COQPATH` in the submakefiles, rather than the parent one.  (I haven't tested it, but I believe it should work, if I didn't make any typos.)

(3) is significantly trickier.  I could have the Makefiles generate `.dir-locals.el`, but I think it would be more trouble than it's worth to make it so that the list of dependencies is literally in only one place, rather than having two lines in each makefile with a comment that is like "if you update either one of these, be sure to update both of them".

Also, I'm not sure how robust the solution to (2) is; I don't know how environment variables are shared between buffers and I think `.dir-locals.el` runs on starting coq-mode for a file, not on starting coqtop on that file, and there is no code that removes things from COQPATH currently.  Perhaps we should discuss in person?

samuelgruetter:
Wow, so you managed to enable support for `COQPATH` without mentioning `COQPATH` in the diff of this PR :joy: :+1:

samuelgruetter:
LGTM, leaving open in case further submodule updating or tweaks are needed

JasonGross:
I've bumped the Kami submodule so that the bundled dependencies are the ones compatible with COQPATH.  Feel free to merge whenever

samuelgruetter:
For the record, here's a fourth desirable property which held before this PR, was lost as of 0ea098a, but restored by 8104c33 (assuming the same changes as done in 8104c33 are also applied to riscv-coq, which I will do soon):

4. If I edit a file in coqutil, and rerun the top-level make, the sub-make in riscv-coq should perform an incremental rebuild based on what changed in coqutil.

With 0ea098a, the `coqdep` running in riscv-coq would not emit dependencies to coqutil (because `coqdep` only seems to emit dependencies to `-Q` and `-R`, but not to files found on `COQPATH`), so the sub-make in riscv-coq doesn't know that it has to rebuild certain files. If one wanted to fix this while sticking to a solution like 0ea098a where during development, sibling projects are included via `COQPATH`, the only fix I can think of would be to run `make clean` in riscv-coq whenever coqutil has changed, so we would not have an incremental build.

PR #107, opened 2019-10-09
standardize mmio strings to "MMIOREAD" and "MMIOWRITE"

andres-erbsen:
untested because the build is [failing](https://gist.github.com/andres-erbsen/0d85ab0dcba161ded69a5808ce906ef6) on my machine


samuelgruetter:
I recently moved some files around and changed the Makefiles, sorry about that, to fix your error you'll have to `rm compiler/_CoqProject bedrock2/_CoqProject end2end/_CoqProject`, and if it still doesn't work, you'll need to `git clean -dfx`.

PR #112, opened 2019-11-05
Update Kami

vbgl:
This is an overlay for https://github.com/coq/coq/pull/11018


samuelgruetter:
Thanks @vbgl! (and sorry that I did not do this myself already, we usually try to only have commits from Kami's `rv32im` branch as a dependency here, but the most recent commits on that branch broke some things in bedrock2, that's why I was waiting...)

vbgl:
Indeed. I’ve updated to the wrong commit. I hope it’s not too bad. Thanks.

PR #113, opened 2019-11-06
parametrize program logic proofs over word impl

andres-erbsen:
remote: Build started: https://builds.sr.ht/~andres/job/105739 [debian-testing.yml]
remote: Build started: https://builds.sr.ht/~andres/job/105740 [nixos-coqmaster.yml]


bedrock2/src/bedrock2Examples/LAN9250.v
```diff
@@ -136,370 +136,441 @@ Require bedrock2Examples.lightbulb_spec.
...
+    rewrite !Z.land_ones by Omega.omega.
+    repeat rewrite Z.mod_mod by (clear; cbv; congruence).
+    rewrite <-!Z.land_ones by Omega.omega.
+    rewrite <-!Z.shiftl_mul_pow2 by Omega.omega.
+    change Semantics.width with 32.
+    set (@word.unsigned _ _ v) as X.
+
+    (* bitwise Z *)
+  
+  Admitted.
```
samuelgruetter: It would be better to only have `case TODO`s and `Qed` instead of `Admitted` to get deeper admit counts. However, I notice that before this PR, this lemma (`lan9250_writeword_ok`) was also admitted, but it does not show up in the final `Print Assumptions`, so it looks like some chain is broken anyways...

samuelgruetter:
70f474d is just me trying out the github conflict resolution UI, we can throw it away if you don't like this direction of merging

bedrock2/src/bedrock2Examples/LAN9250.v
andres-erbsen: `oh there is quite a bit missing between writeword and toplevel, I think the cut right now is lightbulb_init_ok.`

andres-erbsen:
This build has still not yet finished on my machine, I believe it is stuck on https://github.com/mit-plv/bedrock2/pull/113/files#diff-3ea55ef9a278806de25e92151e582c09R177-R178 :/

andres-erbsen:
now it built on my machine with 8.10...

PR #128, opened 2019-12-28
Kami2riscv op immediate

andres-erbsen:


samuelgruetter:
I just pushed an easy fix to make it still work with Coq 8.9.
And I also increased my internal `pain_caused_by_supporting_8.9` counter by 1 :wink: 
At some point (but not yet now) it will reach the limit where I'll suggest to drop Coq 8.9 support.
Travis is green, feel free to merge @andres-erbsen

PR #144, opened 2020-03-14
Generalizes "compile_ext_call_correct" to allow for memory modification

zygi:
Generalizes the "compile_ext_call_correct" interface to allow for memory modification.

Also updates instances of that interface.

### DIFF
--- a/compiler/src/compiler/FlatToRiscvCommon.v
+++ b/compiler/src/compiler/FlatToRiscvCommon.v
@@ -132,7 +132,7 @@ Section WithParameters.
 
     compile_ext_call_correct: forall (initialL: MetricRiscvMachine)
         action postH newPc insts (argvars resvars: list Z) initialMH R Rexec initialRegsH
-        argvals mGive outcome p_sp,
+        argvals mKeep mGive outcome p_sp,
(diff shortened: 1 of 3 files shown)


andres-erbsen:
Cool! (This is Sam's part so he'll have to review the changes.)

Are you also planning to make a riscv `Platform` instance that actually includes memory-modifying external calls? Currently, I believe that the semantics still hardcode lack of memory changes at https://github.com/mit-plv/riscv-coq/blob/fbf15ce26cd019d801dcabb987deeb05970c1731/src/riscv/Platform/MinimalMMIO.v#L180 I think we never quire fleshed out how to fix that. If I recall correctly, we had trouble providing the mmio compiler with the precondition that the io memory is still `None` in the machine state after execution of compiler non-io code, but now that we have `valid_machine`, maybe we could just stick it in there?

zygi:
> Are you also planning to make a riscv Platform instance that actually includes memory-modifying external calls?

Yup! (or extend MinimalMMIO/MMIOSpec, but I assume that's undesirable from the code structure perspective - MMIOSpec looks like a useful interface).

PR #145, opened 2020-03-16
compiler: init sstack list words<->bytes conversion WIP

andres-erbsen:


andres-erbsen:
note to self: an iff1 rewrite fails after I refactored the admitted definition into an existential in the next lemma

andres-erbsen:
merge if builds

samuelgruetter:
Thanks @andres-erbsen :+1:

PR #157, opened 2020-05-05
Adapt to Coq PR #12146: tactic subst now inactive on section variables with indirect dependency in goal

herbelin:
The fix is a priori backward-compatible and can be merged now.

See [coq/coq#12139](https://github.com/coq/coq/issues/12139#issuecomment-622743684) for the motivation as suggested by @JasonGross and supported by @samuelgruetter.

### DIFF
--- a/compiler/src/compiler/RegRename.v
+++ b/compiler/src/compiler/RegRename.v
@@ -724,6 +724,9 @@ Section RegAlloc.
       pose proof E1 as E1'.
       apply @rename_binds_props in E1;
         [|eapply map.empty_injective|eapply map.not_in_range_empty|eapply available_impvars_NoDup].
+      (* remove the indirect dependency of available_impvars in H,
+         and replace H by a dummy hyp to preserve numbering *)
+      clear H. pose I as H.
       simp.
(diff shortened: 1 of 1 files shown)


compiler/src/compiler/RegRename.v
```diff
@@ -870,11 +873,11 @@ Section RegAlloc.
     apply @rename_props in A;
       [|eapply map.empty_injective|eapply map.not_in_range_empty|eapply available_impvars_NoDup].
     specialize (Rrr eq_refl).
-    simp.
+    simp. rewrite Arrrll in *.
```
JasonGross, line 876: Build errors with
```
File "/home/travis/build/mit-plv/bedrock2/compiler/src/compiler/RegRename.v", line 876, characters 10-30:

Error: Nothing to rewrite.
```

herbelin: I forgot to synchronize to the last changes. I did not double-check bedrock2 with Coq master in full but this file passes and it was the only one failing iirc.

samuelgruetter:
Thank you :+1:

PR #162, opened 2020-08-20
Add stackalloc to callees tactic

jadephilipoom:
I was trying out a proof using stack allocation and found that `cmd.stackalloc` is missing from the `callees` tactic, meaning that the goal generated for a function including

```coq
cmd.stackalloc 4 a (cmd.call "foo" [a] ...)
```

ends up without a `spec_of "foo"` instance in scope!


samuelgruetter:
Classic oversight due to a `| _ =>` branch, thanks for fixing!

PR #163, opened 2020-09-12
Merge recent work and update to Coq 8.12

samuelgruetter:
This PR combines some work I did on different branches. It's not finished at all, but doesn't seem to break anything, and doesn't add new admits showing up in the end2end lightbulb theorem, so I'd like to merge it to avoid diverging too much.
Also updates Coq to 8.12.


andres-erbsen:
:+1:

PR #171, opened 2020-12-19
Improving Bytedump.v

herbelin:
This PR puts the two space-only notations for `nil` (`""`) and `Byte.x20` (`"  "`) in different scopes so that they can both work at the same time.

Previously, only the last one was used (i.e. the one for `Byte.x20`) and a special rule for `cons a nil` was compensating the inactiveness of `""`.

Note that this PR is not needed for bedrock2 to work on Coq CI. It is just an update of `Bytedump.v` in passing, taking into account the current status of coq/coq#9741 and coq/coq#9744 that are mentioned in the file.


samuelgruetter:
According to the CI log and the output of `diff -u special/BytedumpTest.golden.hex special/BytedumpTest.out.hex`, the output of `ByteDumpTest.v` is supposed to contain a `0x20` (space), but doesn't any more with this PR. Can you reproduce this failure using `make` inside `bedrock2/bedrock2`?

herbelin:
>  Can you reproduce this failure using make inside bedrock2/bedrock2?

I reproduced. The point is that I tested with a PR I'm working on (coq/coq#13353, for 8.14) and thought that it would work with 8.12 but not. What is missing is that implicit format (in the form of extra spaces in the left-hand side) were not recognized as a new specific format. 

I updated the PR so that it works with 8.12.

Note again that I don't see this Pull "Request" as a "request" to merge. It is more about explaining why it did not work (since there are still open questions in the file) and why two different scopes are (at the end) needed to work.

PR #179, opened 2021-04-15
Hint Mode + for word and map

andres-erbsen:
This would disable type class inference of map and word implementations from guessing the word width, map key type, and map value type. This means that unspecified words or maps fail to typecheck, requiring just enough annotation to reveal how large words or what type of maps are in question. I prepared this change after an annoying debugging session ultimately caused by inference of a locals-to-words map where I had meant to use universally quantified keys. If this is merged, I think 8eb47c3f3101482a5bc17baad80a17fe7116807f could be reverted as well, and a word instance for bytes could be re-added.

The main downside is that several files currently relied on the TC-inferred word instance for `word.of_Z` in a context where the return type was not clear (e.g., `word.unsigned (word.of_Z _)`). I think the ideal solution would look something like `Semantics.word.(of_Z)`, or the equivalent when `Semantics` is imported, but I don't know how to achieve that. There is also annoyance from `replace` not constraining the two sides of the implied equality to have the same type, and a generic fix should be possible for that.

The effect of this patch can be disabled locally using `Local Hint Mode word.word - : typeclass_instances` and `Local Hint Mode map.map - - : typeclass_instances`.

### DIFF
--- a/bedrock2/src/bedrock2Examples/ARPResponder_live.v
+++ b/bedrock2/src/bedrock2Examples/ARPResponder_live.v
@@ -19,39 +19,6 @@ Require Import bedrock2.ZnWords.
 Require Import coqutil.Word.SimplWordExpr.
 Require Import bedrock2.footpr.
 
-(* TODO put into coqutil and also use in lightbulb.v *)
-Module word. Section WithWord.
-  Import ZArith.
-  Local Open Scope Z_scope.
-  Context {width} {word: word.word width} {ok : word.ok word}.
-  Lemma unsigned_of_Z_nowrap x:
-    0 <= x < 2 ^ width -> word.unsigned (word.of_Z (width:=width) x) = x.
-  Proof.
-    intros. rewrite word.unsigned_of_Z. unfold word.wrap. rewrite Z.mod_small; trivial.
-  Qed.
-  Lemma of_Z_inj_small{x y}:
-    word.of_Z x = word.of_Z y :> word -> 0 <= x < 2 ^ width -> 0 <= y < 2 ^ width -> x = y.
-  Proof.
-    intros. apply (f_equal word.unsigned) in H. rewrite ?word.unsigned_of_Z in H.
-    unfold word.wrap in H. rewrite ?Z.mod_small in H by assumption. assumption.
-  Qed.
-
-  Lemma and_bool_to_word: forall (b1 b2: bool),
-    word.and (if b1 then word.of_Z 1 else word.of_Z 0)
-             (if b2 then word.of_Z 1 else word.of_Z 0) =
-    (if (andb b1 b2) then word.of_Z 1 else word.of_Z 0) :> word.
-  Proof.
-    assert (1 < 2 ^ width). {
-      pose proof word.width_pos.
-      change 1 with (2 ^ 0). apply Z.pow_lt_mono_r; blia.
-    }
-    destruct b1; destruct b2; simpl; apply word.unsigned_inj; rewrite word.unsigned_and;
-      unfold word.wrap; rewrite ?unsigned_of_Z_nowrap by blia;
-        rewrite ?Z.land_diag, ?Z.land_0_r, ?Z.land_0_l;
-        apply Z.mod_small; blia.
-  Qed.
-End WithWord. End word.
-
 Module List.
   Import List.ListNotations. Open Scope list_scope.
   Section MapWithIndex.
@@ -1204,16 +1171,6 @@ Ltac ring_simplify_hyp_rec t H :=
 Ltac ring_simplify_hyp H :=
   let t := type of H in ring_simplify_hyp_rec t H.
 
-Lemma if_then_1_else_0_eq_0: forall (b: bool),
-    word.unsigned (if b then word.of_Z 1 else word.of_Z (width:=width) 0) = 0 ->
-    b = false.
-Proof. intros; destruct b; [exfalso|reflexivity]. ZnWords. Qed.
-
-Lemma if_then_1_else_0_neq_0: forall (b: bool),
-    word.unsigned (if b then word.of_Z 1 else word.of_Z (width:=width) 0) <> 0 ->
-    b = true.
-Proof. intros; destruct b; [reflexivity|exfalso]. ZnWords. Qed.
-
 Ltac simpli_getEq t :=
   match t with
   | context[@word.and ?wi ?wo (if ?b1 then _ else _) (if ?b2 then _ else _)] =>
@@ -1226,8 +1183,8 @@ Ltac simpli_getEq t :=
 Ltac simpli :=
   repeat ((rewr simpli_getEq in *  by ZnWords) ||
          match goal with
-         | H: word.unsigned (if ?b then _ else _) = 0 |- _ => apply if_then_1_else_0_eq_0 in H
-         | H: word.unsigned (if ?b then _ else _) <> 0 |- _ => apply if_then_1_else_0_neq_0 in H
+         | H: word.unsigned (if ?b then _ else _) = 0 |- _ => apply word.if_zero in H
+         | H: word.unsigned (if ?b then _ else _) <> 0 |- _ => apply word.if_nonzero in H
          | H: word.eqb ?x ?y = true  |- _ => apply (word.eqb_true  x y) in H
          | H: word.eqb ?x ?y = false |- _ => apply (word.eqb_false x y) in H
          | H: andb ?b1 ?b2 = true |- _ => apply (Bool.andb_true_iff b1 b2) in H
--- a/bedrock2/src/bedrock2Examples/ARPResponder_live.v
+++ b/bedrock2/src/bedrock2Examples/ARPResponder_live.v
@@ -25,12 +25,12 @@ Module word. Section WithWord.
   Local Open Scope Z_scope.
   Context {width} {word: word.word width} {ok : word.ok word}.
   Lemma unsigned_of_Z_nowrap x:
-    0 <= x < 2 ^ width -> word.unsigned (word.of_Z x) = x.
+    0 <= x < 2 ^ width -> word.unsigned (word.of_Z (width:=width) x) = x.
   Proof.
     intros. rewrite word.unsigned_of_Z. unfold word.wrap. rewrite Z.mod_small; trivial.
   Qed.
   Lemma of_Z_inj_small{x y}:
-    word.of_Z x = word.of_Z y -> 0 <= x < 2 ^ width -> 0 <= y < 2 ^ width -> x = y.
+    word.of_Z x = word.of_Z y :> word -> 0 <= x < 2 ^ width -> 0 <= y < 2 ^ width -> x = y.
   Proof.
     intros. apply (f_equal word.unsigned) in H. rewrite ?word.unsigned_of_Z in H.
     unfold word.wrap in H. rewrite ?Z.mod_small in H by assumption. assumption.
@@ -39,7 +39,7 @@ Module word. Section WithWord.
   Lemma and_bool_to_word: forall (b1 b2: bool),
     word.and (if b1 then word.of_Z 1 else word.of_Z 0)
              (if b2 then word.of_Z 1 else word.of_Z 0) =
-    if (andb b1 b2) then word.of_Z 1 else word.of_Z 0.
+    (if (andb b1 b2) then word.of_Z 1 else word.of_Z 0) :> word.
   Proof.
     assert (1 < 2 ^ width). {
       pose proof word.width_pos.
@@ -152,6 +152,8 @@ Section WithParameters.
         (preprocess [autorewrite with rew_word_morphism],
          morphism (Properties.word.ring_morph (word := Semantics.word)),
          constants [Properties.word_cst]).
+
+  Implicit Types m : mem.
 
   (* TODO move (to Scalars.v?) *)
   Lemma load_bounded_Z_of_sep: forall sz addr (value: Z) R m,
@@ -874,7 +876,7 @@ So maybe `P -* P` is equivalent to `emp`? No, because from `P -* P`, `emp` only 
   Goal forall P Q, iff1 Q (sep P (wand P Q)).
   Abort.
 
-  Lemma seps_nth_error_to_head: forall i Ps P,
+  Lemma seps_nth_error_to_head: forall i Ps (P : mem -> Prop),
       List.nth_error Ps i = Some P ->
       iff1 (seps Ps) (sep P (seps (app (firstn i Ps) (tl (skipn i Ps))))).
   Proof.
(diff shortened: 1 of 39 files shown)


JasonGross:
> There is also annoyance from `replace` not constraining the two sides of the implied equality to have the same type, and a generic fix should be possible for that.

```coq
Tactic Notation "myreplace" uconstr(x) "with" uconstr(y) := let ty := open_constr:(_) in replace (x : ty) with (y : ty).
```
or something?

JasonGross:
> I don't know how to achieve that

What's the hard part?

JasonGross:
I don't see an actual `Hint Mode +` nor `Hint Mode !` declaration anywhere, did I miss it?

samuelgruetter:
I can do
```coq
Arguments word.of_Z  {_} _.
Import coqutil.Word.Interface.word.
Check Semantics.word.(of_Z).
(*
of_Z Semantics.word
     : Z -> Semantics.word
*)
```
But it requires importing the identifier `of_Z` unqualified, which is not so nice.

samuelgruetter:
> I don't see an actual `Hint Mode +` nor `Hint Mode !` declaration anywhere, did I miss it?

It's in coqutil, in [this commit](https://github.com/mit-plv/coqutil/commit/7f541a45d9dde1768e4d6ca0d8ff7c224cb11bc3)

JasonGross:
I have no idea if this is what you're looking for, but if you're just looking for name hackery, you could do
```coq
Module word.  Notation of_Z := (word.of_Z (word:=word)). End word.`
```
or something, and then use `Semantics.word.of_Z`?

andres-erbsen:
> I don't see an actual `Hint Mode +` nor `Hint Mode !` declaration anywhere, did I miss it?

in a submodule:
https://github.com/mit-plv/coqutil/compare/6993baf4770dc95f352cce638ab3ebf0e5391a31...7f541a45d9dde1768e4d6ca0d8ff7c224cb11bc3

samuelgruetter:
`Semantics.word` is a getter for a parameter record:
```coq
Semantics.word : forall {parameters : Semantics.parameters}, Interface.word width
```

I guess adding `Module word.  Notation of_Z := (word.of_Z (word:=word)). End word.` in each file that defines a parameter record with a field `word` might work, but I'd prefer a once-and-forall solution.

JasonGross:
> I guess adding `Module word. Notation of_Z := (word.of_Z (word:=word)). End word.` in each file that defines a parameter record with a field `word` might work, but I'd prefer a once-and-forall solution.

This is a once-and-forall solution.  You don't have to do it separately in each file because `Semantics.parameters` is a class with no arguments.  So Coq will happily typeclass resolve whatever instance of `Semantics.parameters` you have lying around if you ask it to resolve `word.of_Z (word:=@Semantics.word _)`, which is what using this notation would do.  You only need to add it in `Semantics.word`.  Then `Semantics.word.of_Z` will typeclass-resolve as you want, and when you `Import Semantics`, you can just use `word.of_Z` (and it'll be the notation defined in `Semantics`).

JasonGross:
Oh, I just realized you were talking about generalizing over `Semantics.parameters`, not over instances of it.  So indeed this is not a once-and-forall solution of the kind you want.

bedrock2/src/bedrock2Examples/ARPResponder_live.v
```diff
@@ -25,12 +25,12 @@ Module word. Section WithWord.
   Local Open Scope Z_scope.
   Context {width} {word: word.word width} {ok : word.ok word}.
   Lemma unsigned_of_Z_nowrap x:
-    0 <= x < 2 ^ width -> word.unsigned (word.of_Z x) = x.
+    0 <= x < 2 ^ width -> word.unsigned (word.of_Z (width:=width) x) = x.
```
andres-erbsen, line 28: duplicate of https://github.com/mit-plv/coqutil/blob/master/src/coqutil/Word/Properties.v#L40 ?

bedrock2/src/bedrock2Examples/ARPResponder_live.v
```diff
@@ -25,12 +25,12 @@ Module word. Section WithWord.
...
   Proof.
     intros. rewrite word.unsigned_of_Z. unfold word.wrap. rewrite Z.mod_small; trivial.
   Qed.
   Lemma of_Z_inj_small{x y}:
-    word.of_Z x = word.of_Z y -> 0 <= x < 2 ^ width -> 0 <= y < 2 ^ width -> x = y.
+    word.of_Z x = word.of_Z y :> word -> 0 <= x < 2 ^ width -> 0 <= y < 2 ^ width -> x = y.
```
andres-erbsen, line 33: duplicate of https://github.com/mit-plv/coqutil/blob/master/src/coqutil/Word/Properties.v#L46 ?

bedrock2/src/bedrock2Examples/ARPResponder_live.v
```diff
@@ -1203,12 +1205,12 @@ Ltac ring_simplify_hyp H :=
   let t := type of H in ring_simplify_hyp_rec t H.
 
 Lemma if_then_1_else_0_eq_0: forall (b: bool),
-    word.unsigned (if b then word.of_Z 1 else word.of_Z 0) = 0 ->
+    word.unsigned (if b then word.of_Z 1 else word.of_Z (width:=width) 0) = 0 ->
```
andres-erbsen, line 1208: https://github.com/mit-plv/coqutil/blob/master/src/coqutil/Word/Properties.v#L392

bedrock2/src/bedrock2Examples/ARPResponder_live.v
```diff
@@ -1203,12 +1205,12 @@ Ltac ring_simplify_hyp H :=
...
     b = false.
 Proof. intros; destruct b; [exfalso|reflexivity]. ZnWords. Qed.
 
 Lemma if_then_1_else_0_neq_0: forall (b: bool),
-    word.unsigned (if b then word.of_Z 1 else word.of_Z 0) <> 0 ->
+    word.unsigned (if b then word.of_Z 1 else word.of_Z (width:=width) 0) <> 0 ->
```
andres-erbsen, line 1213: https://github.com/mit-plv/coqutil/blob/master/src/coqutil/Word/Properties.v#L394

andres-erbsen (commented):
> > There is also annoyance from `replace` not constraining the two sides of the implied equality to have the same type, and a generic fix should be possible for that.
> 
> ```coq
> Tactic Notation "myreplace" uconstr(x) "with" uconstr(y) := let ty := open_constr:(_) in replace (x : ty) with (y : ty).
> ```
> 
> or something?

I meant something like https://github.com/coq/coq/issues/4494#issuecomment-337538890 but sure

andres-erbsen:
@samuelgruetter What about a minimal coercion notation where the word instance name is the prominent part? Something like `Notation "T ! x" := (word.of_Z (word:=T) x) (at level 40).` As we would be able to eventually put this in `word_scope`, using a short character would be fine by me.

bedrock2/src/bedrock2Examples/ARPResponder_live.v
samuelgruetter: yeah probably I moved them to coqutil, and then somehow while rebasing, I accidentally added them again to ARPResponder_live, thx

samuelgruetter:
Having to spell out the name of the instance (usually `word`) is still too long (even though I see that mentioning the instance explicitly helps typechecking), so I'm still advocating for `/[ ... ]` for `word.of_Z` and `\[ ... ]` for `word.unsigned`, but in cases where we must mention the instance, your idea looks good, maybe write it like `Notation "x as T"`, so that it looks like `my_Z_expression as word`?

andres-erbsen:
Okay, it seems that we disagree quite broadly about the presentation of `word.of_Z` (typechecking besides the point).

Would you be willing to accept a variant of this PR where compiler files (and perhaps something else that you care about) import something that sets the typeclass resolution mode back to `-`?

bedrock2/src/bedrock2Examples/ARPResponder_live.v
andres-erbsen: removed in this PR

andres-erbsen:
I tried out a version of Jason's suggestion of overloading `word.of_Z` whenever we overload `word`, and I think it's an improvement over both this PR and current master. I think the effect is exactly as follows: instead of choosing which `of_Z` to use based on which hints are `Required`, the choice is based on which semantics are `Import`ed.

@samuelgruetter, please take another look at the diff against master :). I think I would also be happy to have a shorter name/notation for `Semantics.word.of_Z`, although ideally it would be something that works even when it is not applied to an argument.

compiler/src/compiler/ToplevelLoop.v
```diff
@@ -180,7 +180,7 @@ Section Pipeline1.
 
   Lemma signed_of_Z_small: forall c,
       - 2 ^ 31 <= c < 2 ^ 31 ->
-      word.signed (word.of_Z c) = c.
+      word.signed (word.of_Z (width:=FlatImp.width) c) = c.
```
andres-erbsen, line 183: I was going to lift this to coqutil... I guess I forgot

andres-erbsen: FWIW, the other deleted lemmas worked with `Semantics.word.of_Z`,  but in this one I don't know why FlatImp is the right choice so I didn't feel comfortable changing it.

samuelgruetter:
Asking typeclass search "give me a map instance" without saying which key and value types makes no sense, and therefore adding `Hint Mode +` for maps makes a lot of sense.
However, in all of our current code base, asking typeclass search to "give me a word instance" makes sense, because there's only one kind of words around, namely the words corresponding to the bitwidth of the machine. (And note that I don't consider https://github.com/mit-plv/bedrock2/commit/8eb47c3f3101482a5bc17baad80a17fe7116807f as just a workaround for not knowing how to use `Hint Mode`, but a simplification that makes sense independently because we don't need to perform any operations on non-machine-width words).
So my suggestions would be to always do `Hint Mode +` for maps and never do `Hint Mode +` for words.

andres-erbsen:
Asking for a word instance without width makes no more sense either. Yes, we currently get away with it because we have only convertible instances. But debugging differences between convertible instances still takes time, e g https://github.com/mit-plv/bedrock2/pull/179/files#diff-3b237d67d6d3eae3b6dc756e043d908945be150cdbe1e62415d8291de6a7eb09L466 I think disabling width-guessing typeclass search would be a significant usability improvement for bedrock2 program logic proofs even if we never use multiple (non-convertible) widths.

end2end/src/end2end/End2EndPipeline.v
```diff
@@ -238,7 +238,7 @@ Section Connect.
 
   Lemma riscvMemInit_to_seplog_aux: forall len from,
       Z.of_nat from + Z.of_nat len <= 2 ^ memSizeLg ->
-      PipelineWithRename.ptsto_bytes
+      PipelineWithRename.ptsto_bytes (width:=width)
```
samuelgruetter, line 241: why is this one needed?

processor/src/processor/Consistency.v
```diff
@@ -341,11 +342,11 @@ Section DecExecOk.
       forall w z,
         Z.of_N (wordToN w) = z ->
         krf w =
-        (if Z.eq_dec z 0 then word.of_Z 0
+        (if Z.eq_dec z 0 then word.of_Z (width:=width) 0
          else
            match map.get rrf z with
            | Some x => x
-           | None => word.of_Z 0
+           | None => word.of_Z (width:=width) 0
```
samuelgruetter, line 349: Would it be possible to get rid of the two `width:=width` here by annotating some of the universally quantified variables? (I'm assuming yes, but could you please confirm?)

samuelgruetter:
> Edit: `Semantics.word.of_Z` was precisely my attempt to capture when it makes sense to construct a word without explicitly specifying the width, do you think there are real use-cases not covered by this?

Yeah maybe there aren't, but do we really want to scatter these `Module word. Notation of_Z := (word.of_Z (word:=word)). End word.` everywhere, and have even more word-related puns whose meaning changes when the order of imports changes?

end2end/src/end2end/End2EndPipeline.v
andres-erbsen: I pushed an alternative version that removes some (`*PARAMRECORDS*)` below; fully following the pattern from Semantics.word.of_Z would be to create Utility.word.of_Z. https://github.com/mit-plv/bedrock2/pull/179/commits/0cad84ea9b71852f40816708ec2e9b690906ffc2#diff-075bc0fae1910ae68c5f790e2f170b0b3342a39cc0df6aa9524210bd1a4ca249L241

andres-erbsen:
(I agree that both the boilerplate and order-of-imports issues are real, all I'm saying is that I'd take these issues over syntactic-form-sensitivity of typeclass search any time)

processor/src/processor/Consistency.v
andres-erbsen: Removed.

andres-erbsen:
Here are the word instances available in `End2EndLightbulb` minus `KamiRiscv` which is still building on my end:
```coq
exact word(level 0, pattern Interface.word 32, id 0)
simple apply KamiWord.word(level 0, pattern Interface.word ?META1528, id 0)
simple apply @FlatImp.word(level 0, pattern Interface.word FlatImp.width, id 0)
simple apply @Semantics.word(level 0, pattern Interface.word Semantics.width, id 0)
simple apply @MMIO.MMIO.word(level 1, pattern Interface.word 32, id 0)
simple apply @FE310CSemantics.parameters.word(level 1, pattern Interface.word 32, id 0)
```

compiler/src/compiler/CompilerInvariant.v
```diff
@@ -49,7 +50,7 @@ Section Pipeline1.
   Qed.
 
   Definition imem(code_start code_pastend: Semantics.word)(instrs: list Instruction): Semantics.mem -> Prop :=
-    (ptsto_bytes code_start (instrencode instrs) *
+    (ptsto_bytes (word:=word)(mem:=(@Pipeline.mem p)) code_start (instrencode instrs) *
```
andres-erbsen, line 53: This is an example where the inputs to typeclass search are "known", but their syntactic form leads to a suboptimal resolution.

samuelgruetter: IMHO you're not *solving* the problem here, but just *improving* this specific definition so that the manual PARAMRECORDS related tweak is not needed at (almost) each use site, but only once at definition site. So yes this is an improvement, but I don't think it deserves deletion of the PARAMRECORDS marker, I'd say there should still be a marker at the definition of `imem`. My goal of the PARAMRECORDS markers are to mark annoying snippets of code that would not be needed if we had a solution to this problem, and the above definitely belongs to this category.

andres-erbsen: Fair enough, this entire PR is based on the premise that syntactically distinct but convertible instances are different, whereas a real solution to paramrecords would be some subset of the convertibility relation for which we don't need to worry about this.

PR #180, opened 2021-04-16
Hint Mode + by default, Hint Mode - in files where + required annotations

andres-erbsen:
This is #179 but locally using "Hint Mode -" instead of doing things that Sam thinks would be problematic or ugly. Note that no complete program-logic proof needed net changes.

@samuelgruetter  Can we merge this now so that the work does not go stale, knowing that it is easy to go back if a concrete need or convincing argument emerges? I am also running a build of the ~same changes with `-` as the default, but those have a higher risk of going stale.

### DIFF
--- a/bedrock2/src/bedrock2/AbsintWordToZ.v
+++ b/bedrock2/src/bedrock2/AbsintWordToZ.v
@@ -160,8 +160,6 @@ Module unsigned.
             match reverse goal with H : ?e |- ?G => is_evar e; unify e G; exact H end).. ]
       )) (at level 10, only parsing).
 
-    Lemma absint_of_Z (x : Z) (Hrange : 0 <= x < 2^width) : word.unsigned (word.of_Z (width:=width) x) = x.
-    Proof. rewrite word.unsigned_of_Z; unfold word.wrap; rewrite Z.mod_small; trivial. Qed.
     Definition absint_add (x y : word.rep) ux Hx uy Hy Hbounds : _ =~> _ :=
       absint_lemma! (word.unsigned_add x y).
     Definition absint_sub (x y : word.rep) ux Hx uy Hy Hbounds : word.unsigned _ =~> _ :=
@@ -224,7 +222,7 @@ Module unsigned.
       | _ => match goal with H: word.unsigned e =~> _ |- _ => H end
       | word.of_Z ?a =>
         let Ba := rbounded a in
-        named_pose_proof (absint_of_Z a (boundscheck (X0:=0) (X1:=2^width) Ba (eq_refl true)) : @absint_eq Z (@word.unsigned _ word_parameters e) a)
+        named_pose_proof (word.unsigned_of_Z_nowrap a (boundscheck (X0:=0) (X1:=2^width) Ba (eq_refl true)) : @absint_eq Z (@word.unsigned _ word_parameters e) a)
       | word.and ?a ?b =>
         let Ha := zify_expr a in let Ra := lazymatch type of Ha with _ =~> ?x => x end in
         let Hb := zify_expr b in let Rb := lazymatch type of Hb with _ =~> ?x => x end in
--- a/bedrock2/src/bedrock2/AbsintWordToZ.v
+++ b/bedrock2/src/bedrock2/AbsintWordToZ.v
@@ -160,7 +160,7 @@ Module unsigned.
             match reverse goal with H : ?e |- ?G => is_evar e; unify e G; exact H end).. ]
       )) (at level 10, only parsing).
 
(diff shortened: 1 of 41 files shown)


samuelgruetter:
Looking at `lightbulb_spec.v` still makes me think that `Hint Mode +` for words makes the user experience worse.
But I'd be ok with merging https://github.com/mit-plv/bedrock2/commit/282abfe97ed37ceff61b6211d617e7adb9bc0660

andres-erbsen:
Are you okay with this PR though? That other commit is a proof of concept; I don't really want it even temporarily.

samuelgruetter:
This PR contains some good changes (`Hint Mode +` for maps, `eplace`, moving lemmas to coqutil, maybe more) which we should merge before you run out of steam because of your colleague's disagreements :wink:
It also contains some changes that wouldn't be necessary any more after adding https://github.com/mit-plv/bedrock2/commit/282abfe97ed37ceff61b6211d617e7adb9bc0660, and I don't care too much whether we merge these or not. If existing code contains `(width:=width)` gibberish or the like, that's not a problem per se, it's only a problem if new code will require that gibberish as well, but if we don't do `Hint Mode +` for words, new code will be fine, and I mostly care about how easy it will be to write new code.

andres-erbsen:
Just to be clear, my desired scenario here is that we will not merge 282abfe97ed37ceff61b6211d617e7adb9bc0660. Given this, are you still okay with merging this PR now?

samuelgruetter:
Maybe take another attempt at explaining the advantage of `Hint Mode +` for words? :sweat_smile:  I don't see any at the moment.

andres-erbsen:
Here are the three use cases whose usability I would expect to improve due to fewer statements passing elaboration with unexpected solutions:

1. when factoring out a lemma about word (or memory/locals, which depends on word) from a program logic proof in a file that `Require`s semantics, I have repeatedly thought I have sufficiently parametrized the lemma only to find out that (a) the lemma statement is in fact not true because word expressions were elaborate differently or (b) the seemingly generalized statement did not in fact use the explicitly quantified types, and the proof fails to check when I move it to coqutil. Both cases are strictly a result of user error, it is always possible to write the desired statement. However, code whose word type is only determined by the closest word instance in the context will happily typecheck with a missing quantifier, whereas requiring word type to be specified modulo type inference alone avoids this.
2. I have no mental model for which word instance is chosen by TC search when working in a file that `Require`s multiple word instances, such as End2EndLightbulb.v. I think this includes all concerns you voiced about overloading word by importable aliases, plus some more, because `Require`s are always recursive. Even if the words from these instances are convertible, tactic programming is greatly simplified by having an understanding of the syntactic form, per file if necessary.
3. Algorithms such as chacha20 or Ethernet/IP/UDP checksums are naturally specified in terms of fixed-width, non-64-bit words. We currently don't have any working integrations that include them, but the next planned demo will likely include all three and also some packet formats whose fields are also of smaller width. And saying that I can't specify use coqutil words as types of these fields because we have a policy of only having only one word instance required seems just too capricious.

Now thinking about it, `lightbulb_spec.v` could probably be improved wrt this and other issues by defining the type of traces and type-annotating the predicates themselves.

JasonGross:
> 2\. I have no mental model for which word instance is chosen by TC search when working in a file that `Require`s multiple word instances

https://github.com/coq/coq/pull/14148 (export for instances) might help with this.  My mental model is: if only one instance typechecks, that one is selected, otherwise, I believe hints are ordered by the location at which they are first transitively required and the hint which was transitively required latest is the one that triggers first?  (Or it might be the case that hints are require-scoped but import-ordered, I'm not sure).

samuelgruetter:
> and the hint which was transitively required latest is the one that triggers first

https://coq.inria.fr/refman/proofs/automatic-tactics/auto.html#coq:cmd.Create-HintDb says
> However, the order in which hints are retrieved from the DT may differ from the order in which they were inserted, making this implementation observationally different from the legacy one.

which suggests there are no real guarantees on the order?

(and `theories/Init/Tactics.v` says `Create HintDb typeclass_instances discriminated.` so this DT note does seem to apply).

But anyways, regarding 2), I don't think we can control which typeclass instances appear in our expressions, because if file3 uses file2, which in turn uses file1, file2 might insist on using a parameter transformer instance to turn its parameters into file1parameters, so in file3, even if you try to control which TC instance gets selected, it will still get wrapped in the transformer instance of file2.

andres-erbsen:
We talked about this and decided to merge this as an experiment, leaving open the possibility of cherry-picking 282abfe97ed37ceff61b6211d617e7adb9bc0660 in the future in case we learn we don't want to `Hint Mode +` by default for wrods after all.

PR #181, opened 2021-04-17
Add ProgramLogic.Coercions

andres-erbsen:
```coq
nat >-> Z
word >-> Z
map >-> (map->Prop) (* experimental *)
```

The last one is kinda neat, but will it misfire badly too often? I tried making `sep` take a wrapped version of `map->Prop` as an argument, and that lead to a lot of changes and some unclear design decisions (e.g., what to do with `Lift1Prop` combinators?).

### DIFF
--- a/bedrock2/src/bedrock2/ProgramLogic.v
+++ b/bedrock2/src/bedrock2/ProgramLogic.v
@@ -9,6 +9,20 @@ Require bedrock2.string2ident.
 Definition spec_of (procname:String.string) := list (String.string * (list String.string * list String.string * Syntax.cmd.cmd)) -> Prop.
 Existing Class spec_of.
 
+Module Import Coercions.
+  Import Map.Interface Word.Interface BinInt.
+  Coercion Z.of_nat : nat >-> Z.
+  Coercion word.unsigned : word.rep >-> Z.
+
+  Definition sepclause_of_map {key value map} (m : @map.rep key value map)
+    : map.rep -> Prop := Logic.eq m.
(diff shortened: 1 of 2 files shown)


bedrock2/src/bedrock2/ProgramLogic.v
```diff
@@ -9,6 +9,20 @@ Require bedrock2.string2ident.
 Definition spec_of (procname:String.string) := list (String.string * (list String.string * list String.string * Syntax.cmd.cmd)) -> Prop.
 Existing Class spec_of.
 
+Module Import Coercions.
+  Import Map.Interface Word.Interface BinInt.
+  Coercion Z.of_nat : nat >-> Z.
+  Coercion word.unsigned : word.rep >-> Z.
```
samuelgruetter, line 15: Did you test if this `Coercion word.unsigned : word.rep >-> Z` actually works? (if it does, I'll have to update my understanding of the `Coercion` command or of the uniform inheritance condition)

samuelgruetter:
just to be sure, do coercions follow `Require` or `Import`?

bedrock2/src/bedrock2/ProgramLogic.v
samuelgruetter: Nevermind, I just learnt that the lhs and rhs of `>->` are *classes* (just the head identifier) rather than types, I'd have written
```coq
Coercion word_unsigned_coercion{width: BinInt.Z}{inst: word.word width}(w: @word.rep width inst): BinInt.Z :=
  word.unsigned w.
  ```
instead, but not introducing a new identifier is better.

PR #183, opened 2021-04-18
Demo title Add files via upload

woollybear917:


samuelgruetter:
more cryptocurrency mining, blocked and reported

JasonGross:
Perhaps you want to limit interaction with this repo to people that have had accounts for a while.  That's what GitHub support suggested I do when I complained about getting to much spam commenting on fiat-crypto

andres-erbsen:
I set "Allow local actions only -- Only actions defined in a repository within mit-plv can be used." for now, we'll see.

samuelgruetter:
> Perhaps you want to limit interaction with this repo to people that have had accounts for a while. That's what GitHub support suggested I do when I complained about getting to much spam commenting on fiat-crypto

That's what I already did yesterday, and I posted a message about it in our group chat rather than here because I didn't want to inform the miners about it.

> I set "Allow local actions only -- Only actions defined in a repository within mit-plv can be used." for now, we'll see.

That seems to be the wrong setting. IIUC, an *action* is a standalone github repo that contains nothing more than some CI recipe, like, eg, https://github.com/actions/checkout, or https://github.com/samuelgruetter/check-ci-success, whereas the yml file added by this PR is called a *workflow*, so changing this setting would not have prevented the mining, but makes our fast-forward-script [fail](https://github.com/mit-plv/bedrock2/actions/runs/760233753):

>
> .github/workflows/main.yml : .github#L1
> actions/checkout@v2 and samuelgruetter/check-ci-success@master are not allowed to be used in mit-plv/bedrock2. Actions in this workflow must be: within a repository owned by mit-plv.

So I reverted that settings change.

PR #188, opened 2021-06-04
Use "(void)" if function has no arguments

dayeol:
Function declaration with no argument should generate "(void)"
in order to prevent gcc strict prototype warning (-Wstrict-prototypes)

### DIFF
--- a/bedrock2/src/bedrock2/ToCString.v
+++ b/bedrock2/src/bedrock2/ToCString.v
@@ -116,10 +116,15 @@ Fixpoint c_cmd (indent : string) (c : cmd) : string :=
   end.
 
 Definition fmt_c_decl (rett : string) (args : list String.string) (name : String.string) (retptrs : list String.string) : string :=
-  (rett ++ " " ++ c_fun name ++ "(" ++ concat ", " (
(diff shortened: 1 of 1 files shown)


bedrock2/src/bedrock2/ToCString.v
```diff
@@ -116,10 +116,16 @@ Fixpoint c_cmd (indent : string) (c : cmd) : string :=
...
+    | _ => List.map (fun a => "uintptr_t "++c_var a) args
+    end)
+  in
   (rett ++ " " ++ c_fun name ++ "(" ++ concat ", " (
-                  List.map (fun a => "uintptr_t "++c_var a) args ++
-                  List.map (fun r => "uintptr_t* "++c_var r) retptrs) ++
-                ")").
+  argstring ++
+  List.map (fun r => "uintptr_t* "++c_var r) retptrs) ++
+  ")").
```
samuelgruetter, line 128: What does this output if `args` is `nil` and `retptrs` is non-nil?

dayeol: fixed

dayeol:
sorry, I made a few mistakes so I had to force-push it a few times. Now it is good .

PR #190, opened 2021-07-13
bump coq-record-update

andres-erbsen:


tchajed:
coq-record-update was actually broken. I thought the build failure was only due to the docker image being out-of-date, but actually the test normalization and test output had to be updated. If you bump coq-record-update again bedrock2's CI should now pass.

herbelin:
Sorry for the confusion. Thanks for the portable fix.

PR #191, opened 2021-07-14
Bump coq-record-update to get coq-record-update PR #23 (for Coq PR #14596)

herbelin:
For compatibility of Coq CI with bedrock2, I forgot to update the HEAD of the coq-record-update submodule. This PR does the update.

Assuming the procedure is ok, this should be merged as soon as possible to preserve the compatibility of Coq CI works with bedrock2. Thanks in advance.


herbelin:
Oops, there is already #190. Closing.

PR #194, opened 2021-07-21
Array lemmas

bdiehs:
Created lemmas for reasoning about program logics involving array loads and stores.


samuelgruetter:
* At first sight, I found it a bit surprising that these lemmas take both a `size: word` and an `sz: access_size` that are not related in any way, but since you can prove the lemmas without any hypotheses relating them, that should be fine (that means the lemma applies to contradictory cases where you have overlapping words (`size < sz`) and to cases where the array contains gaps between the elements (`size > sz`), but I guess it doesn't hurt also accidentally cover cases that are unlikely to be used)
* I'm not sure how likely it is that these lemmas would be used for `sz=access_size.one`, because then you'd probably use an array of bytes rather than an array of words of which only the lowest byte matters. But maybe for `sz=access_size.two` one might want to use an array of words, even though the upper bytes would be ignored. And given that these lemmas hold for any `sz`, I think it's good to prove them for all `sz`, even if they might not get used for all `sz`.
* There are a bit too many merge and revert commits in this PR, but the overall diff looks good, so I'd suggest that I just use Github's  "Squash and merge" feature to turn it into one single commit.

Overall, LGTM (and I hope we won't get too much opposition from @andres-erbsen who at some point expressed dislike for array lemmas, IIRC :wink:)

andres-erbsen (approved):
Agreed with everything that Sam says. If my different aspirations for memory reasoning pan out we might reconsider the usefulness of these lemmas, but even then I imagine there being use cases where we want them.

And yes, squashing is required here.

PR #200, opened 2021-10-04
Ecancel impl

DIJamner:
Generalizes `ecancel` to work on `impl1` in addition to `iff1` and modifies `ecancel_assumption` to use the new code path. Currently leads to some performance regression, on the order of 10% for the full development on my machine.

### DIFF
--- a/bedrock2/src/bedrock2/Map/SeparationLogic.v
+++ b/bedrock2/src/bedrock2/Map/SeparationLogic.v
@@ -460,7 +460,8 @@ Ltac ecancel_step :=
    but that is expensive
  *)
 Ltac is_evar_or_alias e :=
-  is_evar e || (let e' := eval unfold e in e in is_evar e').
+  tryif is_evar e then idtac
+  else let e' := eval unfold e in e in is_evar e'.
 
 (* TODO: eventually replace ecancel_step? Probably not since implication is too agressive.*)
 (*TODO: performance. I've replaced the deltavar stuff with heavier operations  *)
@@ -474,7 +475,7 @@ Ltac ecancel_step_by_implication :=
          Makes sure j doesn't point to an evar
        *)
       (*TODO: need to check for bound idents too; use above fn*)
-      assert_fails (idtac; is_evar y);      
+      assert_fails (idtac; is_evar_or_alias y);      
       let LHS := lazymatch goal with |- Lift1Prop.impl1 (seps ?LHS) _ => LHS end in
       let i := find_implication LHS y in (* <-- multi-success! *)
       (*TODO: nocore?*)
@@ -512,14 +513,9 @@ Ltac cancel :=
   repeat cancel_step;
   repeat cancel_emp_l;
   repeat cancel_emp_r;
-  try solve [ cancel_done ].
-
-
-Ltac cancel_impl :=
-  reify_goal;
-  repeat cancel_step;
   repeat cancel_emp_impl;
   try solve [ cancel_done ].
+
 
 (*TODO: should I use deltavar here too?*)
 (*TODO: this assumes that the goal has (at most or exactly?) 1 evar predicate.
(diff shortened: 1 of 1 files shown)


bedrock2/src/bedrock2/Map/SeparationLogic.v
```diff
@@ -388,6 +454,33 @@ Ltac ecancel_step :=
...
+      let j := lazymatch jy with (?i, _) => i end in
+      let y := lazymatch jy with (_, ?y) => y end in
+      (* For implication, we don't want to touch RHS evars 
+         until everything else is solved.
+         Makes sure j doesn't point to an evar
+       *)
+      (*TODO: need to check for bound idents too; use above fn*)
+      assert_fails (idtac; is_evar_or_alias y);      
+      let LHS := lazymatch goal with |- Lift1Prop.impl1 (seps ?LHS) _ => LHS end in
+      let i := find_implication LHS y in (* <-- multi-success! *)
```
samuelgruetter, line 480: It seems that compared to the original `ecancel_step`, where `find_syntactic_unify_deltavar` only returns lhs clauses that syntactically match the rhs clause, you now increase the number of returned lhs clauses to *always return all of them*. This seems quite expensive, I'd prefer this not to be the default.

bedrock2/src/bedrock2/Map/SeparationLogic.v
```diff
@@ -368,6 +403,37 @@ Ltac cancel_seps_at_indices i j :=
...
+    simple refine (cancel_seps_at_indices_by_implication i j LHS RHS _ _);
+    cbn [firstn skipn app hd tl]
+  end.
+
+
+(*TODO: use hintdb or user-provided tactic?
+  Study performance implications
+*)
+Create HintDb ecancel_impl discriminated.
+Hint Extern 1 => exact (fun m x => x) : ecancel_impl.
```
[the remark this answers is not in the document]
DIJamner: Yes, that should be fine. I'll make that change.

bedrock2/src/bedrock2/Map/SeparationLogic.v
```diff
@@ -368,6 +403,37 @@ Ltac cancel_seps_at_indices i j :=
...
+
+(*TODO: performance*)
+Ltac find_implication xs y :=
+  multimatch xs with
+  | cons ?x _ =>
+    let H := fresh in
```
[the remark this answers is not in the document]
DIJamner: Thanks for catching that, it's dead code.

bedrock2/src/bedrock2/Map/SeparationLogic.v
DIJamner: Yes, that's currently how it works. I figured that, since for each pair I now need to search the available hints, I would rather not do this search twice. It might be the case that this would be faster than what I have if the code between here and where backtracking occurs is expensive enough, but I haven't tried the other way. I could also try to pass the generated proof out along with i, but that will require a bit more work.

bedrock2/src/bedrock2/Map/SeparationLogic.v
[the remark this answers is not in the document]
DIJamner: It turns out that it's better if this hint is at level 0, since it should always take priority over other hints.

bedrock2/src/bedrock2/Map/SeparationLogic.v
```diff
@@ -420,12 +502,21 @@ Ltac cancel :=
...
 
+
 Ltac ecancel :=
   cancel;
-  repeat ecancel_step;
-  solve [ ecancel_done ].
+  lazymatch goal with
+  | [|- impl1 _ _] =>
+     repeat ecancel_step_by_implication;
+     (solve [ cbn [seps]; exact (fun m x => x) ])
```
samuelgruetter, line 514: Could this be `exact impl1_refl`? For readability, I don't quite like throwing around untyped proof terms...

bedrock2/src/bedrock2/Map/SeparationLogic.v
```diff
@@ -420,12 +498,21 @@ Ltac cancel :=
   repeat cancel_step;
   repeat cancel_emp_l;
   repeat cancel_emp_r;
+  repeat cancel_emp_impl;
```
samuelgruetter, line 501: Quick question: does this only cancel `emp`s on the right, but not on the left? If so, is that intended?

samuelgruetter: So what does your new `ecancel` do on `impl1 (emp True * emp True * emp True) (emp True)`?

DIJamner: It doesn't currently handle that. The examples I was considering typically had an evar `?R` on the RHS and extra `emp` clauses would just get folded into that. You can always discharge `emp P` on the LHS, so I think I'll add a pass at the end to remove any that are left over.

PR #207, opened 2021-12-11
clean up NotationsCustomEntry printing

andres-erbsen:


bedrock2/src/bedrock2Examples/Demos.v
```diff
@@ -4,15 +4,12 @@ Require Import bedrock2.NotationsCustomEntry.
...
-Local Coercion expr.literal : Z >-> expr.
-Local Coercion expr.var : String.string >-> expr.
-Local Coercion name_of_func (f : function) : String.string := fst f.
+Import BinInt String bedrock2.Syntax.Coercions.
 
 Definition bsearch: func :=
   let left := "left" in let right := "right" in let target := "target" in let mid := "mid" in
   ("bsearch", ([left; right; target], [left], bedrock_func_body:(
-  while (right - left) {
+  for (right - left) {
```
samuelgruetter, line 12: Why do you say `for` instead of `while`?

bedrock2/src/bedrock2Examples/ArrayLoadStore.v
```diff
@@ -31,9 +28,9 @@ Section WithParameters.
       let j := "j" in
       let r := "r" in
     ("tf", ([buf; len; i; j], [], bedrock_func_body:(
-      require ( i < len ) else { /*skip*/ };
+      match ( i < len ) else { /*skip*/ };
```
samuelgruetter, line 31: Why `match`?

bedrock2/src/bedrock2Examples/stackalloc.v
```diff
@@ -4,21 +4,22 @@ Import Syntax.Coercions BinInt String List.ListNotations.
...
+  ("stacktrivial", ([]:list String.string, [], bedrock_func_body:(stackalloc! 4 as t ; /*skip*/ ))).
 
 Definition stacknondet : bedrock_func := let a := "a" in let b := "b" in let t := "t" in
-  ("stacknondet", ([]:list String.string, [a; b], bedrock_func_body:(stackalloc 4 as t {
+  ("stacknondet", ([]:list String.string, [a; b], bedrock_func_body:(
+  stackalloc! 4 as t;
   a = (load4(t) >> coq:(8));
   store1(a+coq:(3), coq:(42));
   b = (load4(t) >> coq:(8))
-}))).
```
samuelgruetter, line 14: If you remove curly braces, you lose the ability to have two stackallock blocks in sequence, where the second reuses the stack space of the first. And if you put a stackallock both inside a then branch and inside an else branch, you'll be surprised that the stackalloc'd memory is not available any more after the if.

bedrock2/src/bedrock2/NotationsCustomEntry.v
```diff
@@ -153,35 +156,45 @@ Notation "f args" :=  (call nil f args) (in custom bedrock_cmd at level 0, f glo
...
 
 Module test.
-  Local Notation "1" := (expr.literal 1) (in custom bedrock_expr).
+  Local Notation "1" := (expr.literal 1) (in custom bedrock_expr at level 0).
 
   Goal True.
+
+  pose (fun while : nat => while=while).
+  pose (fun stackalloc : nat => stackalloc=stackalloc).
+  pose (fun require : nat => require=require).
```
andres-erbsen, line 192: @samuelgruetter i switched to using existing keywords because "while" is already used as an identifier downstream

bedrock2/src/bedrock2Examples/stackalloc.v
andres-erbsen: you can still upt curly braces around stackalloc.

the scoping discrepency is indeed annoying but it's already present in the semantics -- the stackalloc name overwrites instead of shadowing. i'm not sure which way is worse.

samuelgruetter:
> because "while" is already used as an identifier downstream

Could we adhere to the following principle, and would that allow us to keep using `while`?: *If a file wants to `Import NotationsCustomEntry`, it must not use `while` as an identifier, but if it only (potentially transitively) `Require`s `NotationsCustomEntry`, using `while` as an identifier is ok*.

The following snippet suggests that as long as you don't import the notations, it's fine:

```coq
Inductive cmd: Type := SSkip | SWhile(b: bool).

Module MyNotations.
  Notation "'while' ( b ) { }" := (SWhile b).
  Check (while (true) {}).
End MyNotations.

Definition foo(while: nat) := while. (* works *)

Import MyNotations.

Definition bar(while: nat) := while. (* parsing error *)
```

Or can you produce a similar minimal example that shows why you want to use only existing keywords?

andres-erbsen:
i'll try that out. based on experience with earlier versions i didn't expect what you posted to work.

downstream code does mix gallina definitions that look like imperative code with bedrock2 snippets though. would you be fine with providing threw notations i suggest now *and* the previous ones, with the previous ones needing an extra import but taking prrecedence if imported?

samuelgruetter:
Using `for` instead of `while` and `match` instead of `require` is really confusing, and makes presenting our source code an embarrassing experience, I'd like to avoid that.

> downstream code does mix gallina definitions that look like imperative code with bedrock2 snippets though.

Could you give an example?

andres-erbsen:
Having importing our code break other working code is an embarrassing experience and I would like to avoid that more than using constrained keywords. `for` for `while` is the standard syntax for loops in Go, I don't think there's anything bad about that. `match` is indeed unfortunate, but it's also for a weird feature that I don't expect to go without explaining with any keyword. I really don't believe we are losing much here.

I did check that we could make `while ` keyword guarded by Import as you suggest, so if that's the way that lets us move forward, I can revert to that. However, having these notations not imported means that printing bedrock2 code will show lisp-style ASTs instead, which surely isn't better -- and I don't think I can recommend importing something that makes standard programming terminology into keywords with a straight face.

Some examples that use both inline bedrock2 parsing/printing and gallina models (but these are not examples of conflicts, at least for parsing, just contexts where consciously avoiding conflicts would cause programmer burden):
https://github.com/mit-plv/rupicola/blob/8483582b178dba127ee4f9335aa1dc3befc79b1d/src/Rupicola/Lib/ControlFlow/DownTo.v#L377-L452
https://github.com/mit-plv/fiat-crypto/blob/f08e5c5a1ceda7aea538820821714fb0df7d59c4/src/Bedrock/Group/ScalarMult/MontgomeryLadder.v

andres-erbsen:
I opened https://github.com/coq/coq/issues/15349

As we are brainstorming notations already, what do you think of writing `$1234` for `coq:(1234)`? More confusing, less of an eyesore? Or maybe I should just try to resurrect https://github.com/coq/coq/pull/11561...

samuelgruetter:
> `for` for `while` is the standard syntax for loops in Go

Aah now I see why you don't consider this waay beyond reasonable like me who have never seen Go loops before :sweat_smile: But still, we keep claiming to be somewhat close to C, and exportable to C, and many people haven't used Go, so it would really be better to avoid using `for` for `while`.

`Require Foo` should not have unexpected side effects, but `Require Import Foo` should be allowed to have side effects on many things, in particular, it can already change the meaning of an identifier that previously referred to something else. But I also see that you'd like to pretty-print ASTs whenever possible...

Another hack to consider might be to make `while(` a keyword?

And for literals, I recently briefly tried out `Number Notation`:

```coq
Definition print_literal(e: Syntax.expr): option Z := 
  match e with
  | expr.literal z => Some z
  | _ => None
  end.
Declare Scope literal_expr_scope.
Delimit Scope literal_expr_scope with literal_expr.
Number Notation Syntax.expr expr.literal print_literal : literal_expr_scope.
Goal 42%literal_expr = expr.literal 42. reflexivity. Abort.
```

I ended up not using it because I could only have one default rule that had to work both for variables (that I wanted to parse with an ident-in-context-to-string Ltac hack) and literals, but if you don't need this hack, `Number Notation` might be worth a try.

andres-erbsen:
I think numeral notations are blocked on the coq PR I linked above, no?

(I'm still not sure what to do about keywords: breaking parsing seems much worse than changing the meaning of a keyword, and I worry `while(` would still trigger that...)

Yes, the precedences are already as intended. It's far from controversial that operator precedence in C has mistakes, with the clearest case being https://www.lysator.liu.se/c/dmr-on-or.html 

More generally, I think there is little to be gained from being (nearly) source-compatible with C given the degree to which expected workflow here differs from usual usage of C. If a potential user is scared away by syntactic differences (of the degree as found between existing widely-used languages, which perhaps does not include the `coq:( .. )` mess), I don't think there is much we can do when they get to proofs.

samuelgruetter:
> I think numeral notations are blocked on the coq PR I linked above, no?

Ah yes, I guess at the moment, the only thing that a `Numeral Notation` would buy you would be to not have to make `expr.literal` a coercion.
But actually, have you tried replacing `Notation "'coq' : ( e )" := e (e constr, ...)` by simply `Notation "e" := e (e constr, ...)`?

> More generally, I think there is little to be gained from being (nearly) source-compatible with C 

Yes if you use `ToCString`. (The reason I followed exactly the C precedence is that I still hope to be able to do something like `echo '/*' > out.c && cat my_live_verif_file.v >> out.c && echo '*/' >> out.c && gcc out.c` one day.)

Another idea: What about replacing `require`/`match` by something like this:

```coq
Notation "'if' ( e ) { c1 ; 'return' ; } c2" := (cmd.cond e c1 c2) (at level 0).
```
(combined with some machinery to make sure this is only used at top-level).

And then, as per https://github.com/coq/coq/issues/15349#issuecomment-992879786, just accept that parsing might get broken by importing notations in 8.14, but not any more in 8.15?

andres-erbsen:
Oh, all good ideas, let me try!

andres-erbsen:
I think `stackalloc` at level 0 (to keep it an identifier) does not work because it should bind less tightly than the `;` that follow it, and `;` needs to be above level 0 to be printed correctly. Your suggestion with `return` is probably next up as `return` is already a keyword.

bedrock2/src/bedrock2/NotationsCustomEntry.v
```diff
@@ -37,151 +34,127 @@ Infix "=="  := (expr.op  eq)  (in custom bedrock_expr at level 10, no associativ
...
+Notation "x" := x (in custom bedrock_cmd, x global).
+
+(* must be at level 1 or higher to arguments make sense, see coq/coq#15336 *)
+Notation "c1 ; c2" := (seq c1%bedrock_nontail c2)
+  (in custom bedrock_cmd at level 1, right associativity, format "'[v' c1 ; '/' c2 ']'").
+
+(* the notations that start with a word must be at level 0 to avoid making that
+   word into a global keyword, see coq/coq#15349 coq/coq#14562 *)
+Notation "'while' e { c }" := (while e c%bedrock_nontail)
+  (in custom bedrock_cmd, e custom bedrock_expr, format "'[v' 'while'  e  {  '/  ' c '/' } ']'").
```
samuelgruetter, line 63: If the comment says "must be at level 0", the code should also explicitly say `at level 0` rather than relying on this being the default.

andres-erbsen:
I think `constr` conflicts with `bedrock_cmd` and `bedrock_expr` too badly to be mixed, for example:

```coq
Declare Custom Entry bedrock_expr.
Notation "e"          := e (in custom bedrock_expr at level 0 , e constr).
Declare Custom Entry bedrock_cmd.
Notation "bedrock_cmd:( c )" := c (c custom bedrock_cmd, format "'bedrock_cmd:(' '/  ' c '/ ' ')'").
Notation "/*skip*/" := cmd.skip (in custom bedrock_cmd).
Notation "'if' e { c1 }"               := (cmd.cond e c1 cmd.skip)  (in custom bedrock_cmd at level 0,
  e custom bedrock_expr at level 1, format "'[v' 'if'  e  {  '/  ' c1 '/' } ']'").
Goal True.
  epose (bedrock_cmd:( /*skip*/ )).
  epose (bedrock_cmd:(
  if 1 { /*skip*/ } else if 1 { /*skip*/ } )).
```

The `if`/`return` idea seems cute, but I don't know for the top of my head how to allow multiple commands before return there, and since `require` seems to work again, I'm not sure whether to pursue it.

bedrock2/src/bedrock2/NotationsCustomEntry.v
andres-erbsen: I don't see the case here. Other notations in this file also rely on the implicit level assignment, and I don't expect changing their level to be any less damaging -- the comment just highlights a non-obvious constraint.

andres-erbsen:
ideally if-then-else-if should be formatted with the second if body at the same indent as the first, but I don't seem to be able to get it right right now.

samuelgruetter (approved):

JasonGross:
For multiple commands before return, does something like this work?
```coq
Notation "'if' ( e ) { c10 ; .. ; c11  ; 'return' ; } c2" := (cmd.cond e (cons c10 .. (cons c11 nil) .. ) c2) (at level 0).
```
(Replacing cons with whatever sequencing you want)

PR #209, opened 2021-12-14
NotationsCustomEntry: $(constr)

andres-erbsen:


samuelgruetter:
As far as I understand, we would like to add two rules of the form

```coq
Notation "x" := x (in custom expr, ...)
```

where one is for variables, and the other for number literals. But Coq's notation system only allows one such identity notation, so we need to use a symbol (such as `$`, for instance) for the other one.

Here you decided to keep using the identity notation for variables, and start using `$` for numbers, but how about the other way round? Combined with `ident_to_string`, we could say that all variables have to start with `$`, but then we wouldn't need quotes any more.

Here's a quick experiment (relying on `bedrock2Examples.LiveVerif.ident_to_string`, which you can feel free to edit/move if you want):

```coq
Require Import Coq.ZArith.ZArith. Open Scope Z_scope.
Require Import Coq.Strings.String.
Require Import bedrock2Examples.LiveVerif.ident_to_string.

Inductive expr :=
| EAdd: expr -> expr -> expr
| EVar: string -> expr
| ELit: Z -> expr.

Coercion ELit: Z >-> expr.

Declare Custom Entry expr.

Infix "+" := EAdd (in custom expr at level 5).

Notation "x" := x (in custom expr at level 0, x bigint).
Notation "$ x" := (EVar (ident_to_string x))
  (in custom expr at level 0, x constr at level 0, only parsing).

Notation "'q:(' x )" := x (x custom expr, format "'q:(' x )").

Goal q:(3 + $y) = EAdd (ELit 3) (EVar "y"). reflexivity. Abort.
```

I still haven't figured out how to print `$x`, but I believe this should be possible.
For me, either way of using `$`/the identify notation for variables/literals or the other way round is fine, just wanted to make sure the choice is conscious :wink:

andres-erbsen:
> where one is for variables, and the other for number literals. But Coq's notation system only allows one such identity notation

As I understand it, it does not allow any. Is there a way?

andres-erbsen:
Oh, maybe it is one, and we are already using `global`?

samuelgruetter:
> Would you mind if we renamed `ident_to_string` to `ident_to_string!`?

Good point, I like the convention to end notations that invoke Ltac with an exclamation mark.

And currently, we're using an "identity notation" (what you call "grammar coercion" I guess?) with `global`:

https://github.com/mit-plv/bedrock2/blob/84ccc38a7244a453155023ceadc975ad4c65d9b7/bedrock2/src/bedrock2/NotationsCustomEntry.v#L11

but actually, we only rely on it to refer to let-bound identifiers whose right-hand side is a string with the same name as the identifier, is that right?
Why are we using `global` and not `name` then?

Also, I find the name `global` confusing, because it also allows let-bound identifiers:

```coq
Require Import Coq.ZArith.ZArith. Open Scope Z_scope.
Require Import Coq.Strings.String. Open Scope string_scope.

Module m.
  Definition foo: string := "foo".
End m.

Inductive expr :=
| EAdd: expr -> expr -> expr
| EVar: string -> expr
| ELit: Z -> expr.

Coercion EVar: string >-> expr.

Declare Custom Entry expr.

Infix "+" := EAdd (in custom expr at level 5).

Notation "x" := x (in custom expr at level 0, x global).
Notation "$ x" := (ELit x)
  (in custom expr at level 0, x constr at level 0, format "$ x").

Notation "'q:(' x )" := x (x custom expr, format "'q:(' x )").

Goal let y := "y" in q:(y + $3 + m.foo) = EAdd (EAdd (EVar y) (ELit 3)) (EVar m.foo).
reflexivity. Abort.
```

andres-erbsen:
We also use `global` refer to definitions like `MMIO_SPI_ADDR`; these will have to be `$`-prefixed if idents are used as bedrock2 identifiers.

bedrock2/src/bedrock2Examples/FE310CompilerDemo.v
```diff
@@ -75,30 +73,30 @@ Definition swap_chars_over_uart: cmd :=
 
   bedrock_func_body:(
     (* ring oscillator: enable, trim to 72MHZ using value from OTP, divider=0+1 *)
-    io! rx = MMIOREAD (coq:(Ox"0x00021fec"));
-    output! MMIOWRITE(hfrosccfg, coq:(2^30) | (rx & coq:(2^5-1)) << coq:(16));
+    io! rx = $MMIOREAD (coq:(Ox"0x00021fec"));
+    output! $MMIOWRITE($hfrosccfg, coq:(2^30) | (rx & coq:(2^5-1)) << $16);
```
samuelgruetter, line 77: How about deleting these two definitions:

https://github.com/mit-plv/bedrock2/blob/84ccc38a7244a453155023ceadc975ad4c65d9b7/bedrock2/src/bedrock2/FE310CSemantics.v#L12-L13

And then write `MMIOREAD`/`MMIOWRITE` without `$`, but using `ident_to_string`?

Also, how do function calls work now? Are you still using a coercion to project the first field (function name) out of the function tuple, or is it also `ident_to_string` based?

andres-erbsen: I used ident to string everywhere for now, `global` has https://github.com/coq/coq/issues/15360 :shrug:

andres-erbsen:
@samuelgruetter did you evaluate https://github.com/mit-plv/koika/blob/37a4d8b1fe9b4b9a131e193281ebd33b28ce1c35/coq/IdentParsing.v#L300-L331 ? I wonder whether it would help with the `$_` issue (https://github.com/mit-plv/bedrock2/pull/209#issuecomment-993979984).

samuelgruetter:
I haven't actually used `ident_to_string` yet, I've only used its building block `exact_varconstr_as_string` (to convert bound identifiers to strings), and also haven't looked into Koika's typeclass-based approach. Does it solve your `$_` issue?

samuelgruetter:
Ah no, correction: I did use `ident_to_str`, but only on the LHS of assignments, and the tactics-in-terms approach worked for me, but we can also switch to a typeclass hack if it solves a problem for you.

andres-erbsen:
771b04d5cfa437d53ffe75c76d915e5858456841

PR #212, opened 2021-12-15
Update .gitmodules with kami branch rv32i

JasonGross:
So that dependabot knows which branch to use


andres-erbsen:
@joonwonc do you remember whether we should be tracking this branch?

joonwonc:
If you're asking which Kami branch Bedrock2 has used so far, yes `rv32i` is the one.

andres-erbsen:
Yes, thank you!

PR #213, opened 2021-12-16
fix `ToCString.c_bop` type, `bytedump.sh`, test

andres-erbsen:
ToCString.expr should always generate C code whose C type is `uintptr_t`, but this wasn't actually true for all binary operators. If these operators where used as subexpressions (instead of rhs of an assignment) the downstream operations would be inferred with other sizes, computing incorrect values.

This PR fixes this on a case-by-case basis and adds a test to make sure all expressions have a type compatible with `uintptr_t`. Without the fixes, the test failed for `mulhuu`, `.>>`, `<`, `.<`, and `==` -- exactly the operators we identified as problematic by reading the specification.

Thanks to @cpitclaudel for pointing out this issue!

I took the opportunity to refactor `bytedump.sh` and the existing `Bytedump` test to work around https://github.com/coq/coq/issues/15373 and reused them for the C test.

The 32-bit test does not pass (at least on Arch Linux and gcc) because `__uint128_t` is unavailable. Since this is error is compile-time and trivially-workaroundable, I'll postpone looking into this until we have a 32-bit use case for this backend.


cpitclaudel:
:+1: Are you planning to do a similar PR for the alloca memory leak?

cpitclaudel:
(And, any chance you could apply this fix to the vendored copy in Rupicola as well?)

bedrock2/Makefile.coqflags
```diff
@@ -0,0 +1,4 @@
+.PHONY: all print_coqc
+-include Makefile.coq.ex.conf
+print_coqc:
+	@printf "%s %s\n" "$(COQMF_COQLIBS)" "$(COQMF_OTHERFLAGS)"
```
samuelgruetter, line 4: When does this file come into play?

bedrock2/Makefile
```diff
@@ -62,7 +64,14 @@ special/BytedumpTest.out: special/BytedumpTest.v special/BytedumpTest.golden.bin
 	diff -u special/BytedumpTest.golden.bin special/BytedumpTest.out.tmp
 	mv special/BytedumpTest.out.tmp special/BytedumpTest.out
 
-force:
```
samuelgruetter, line 65: Did you delete this line on purpose?

bedrock2/Makefile.coqflags
andres-erbsen: I use it to run `bytedump.sh` from the command line.

bedrock2/Makefile
andres-erbsen: No. What does it do?

samuelgruetter: A target without prerequisites and without recipe is considered always outdated, and by making `Makefile.coq.ex` and `Makefile.coq.noex` depend on it, we can force these to be regenerated each time, so that when we add or delete `.v` files, they automatically get added to/removed from the Makefile. I guess since `force` is phony, generating `Makefile.coq.ex` and `Makefile.coq.noex` still worked on this PR even though no `force` target exists any more, but could you check if `coq_makefile` is also still re-run on each `make` invocation?

bedrock2/Makefile.coqflags
samuelgruetter: Does the Makefile or any checked-in file contain a command that depends on this `Makefile.coqflags`? If not, it would be good to add a comment as to why this file is here.

PR #217, opened 2021-12-16
ToCSTring: use array declaration instead of alloca, test

andres-erbsen:
From https://github.com/mit-plv/bedrock2/pull/213#issuecomment-995458380 and https://github.com/mit-plv/bedrock2/pull/213#discussion_r770537960. I confirmed that the test segfaults without the change.


bedrock2/Makefile.coqflags
```diff
@@ -1,6 +1,9 @@
+# This file can be used to print the flags Makefile.coq.ex pass
+# Example: coqtop $(make -f Makefile.coqflags)
```
samuelgruetter, line 2: ah cool, that looks useful :+1:

PR #219, opened 2021-12-16
replace LittleEndian with LittleEndianList in bedrock2/

andres-erbsen:
I replaced LittleEndian with LittleEndianList in `bedrock2/`, most notably in `Scalars.v`. I did not change any high-profile definitions from tuples to lists yet. Because of this, a few new tuple-list conversions appear, and the length parameters of some tuples change from `n` to `length (le_split n v)` in a way that is inconvenient in some places. I think these would go away when `ptsto_bytes` starts taking a list instead of a tuple, but I do not dare to change that much at once. Being able to cancel out `tuple.of_list` and `tuple.to_list` instead of doing dependent-type hackery is quite satisfying, though.

I did the porting up to `FlatToRiscvCommon` where a lemma's arguments are have types dependent on an `access_size`. I don't think that place is actually stuck, but I did start wondering whether a less local solution might be nicer, and now I'm getting distracted. The fallback would be something like https://github.com/mit-plv/bedrock2/compare/LittleEndianList-transition?expand=1#diff-ecabcc922edaeb5d11aca7cb817e8a72dc43591edfe439ddd7b0d16d90b568ffR61

@samuelgruetter would you be willing to give a try to pushing this change through the compiler? I'll probably continue up towards rupicola and fiat-crypto next time I work on code, which might be notably into next week.

Here is the list of remaining occurrances of `LittleEndian`. it is not an exhaustive list of what might break, but I think I got a good sense from the half of this list that I did deal with. Also note that I am not rushing to remove `LittleEndian` from everywhere, and what we need to do to get this branch to build is largely just accounting for ripples for `Scalars.v`.

```coq
compiler/src/compiler/FlatToRiscvCommon.v:      rewrite LittleEndian.combine_split.
compiler/src/compiler/FlatToRiscvCommon.v:    rewrite bitSlice_all_nonneg. 2: cbv; discriminate. 2: apply (@LittleEndian.combine_bound 4).
compiler/src/compiler/FlatToRiscvCommon.v:    rewrite LittleEndian.split_combine.
compiler/src/compiler/FlatToRiscvCommon.v:    apply (@LittleEndian.combine_bound 4).
compiler/src/compiler/Softmul.v:    rewrite LittleEndian.combine_split. rewrite word.unsigned_of_Z_nowrap. 2: apply encode_range.
compiler/src/compiler/RunInstruction.v:                  (word.of_Z (opt_sign_extender (LittleEndian.combine n v))) /\
compiler/src/compiler/RunInstruction.v:        (Exec * ptsto_bytes n addr (LittleEndian.split n (word.unsigned v_new)) * R)%sep
compiler/src/compiler/CompilerInvariant.v:      HList.tuple.to_list (LittleEndian.split 4 (encode instr)) ++ instrencode instrs.
compiler/src/compiler/GoFlatToRiscv.v:      rewrite LittleEndian.combine_eq, HList.tuple.to_list_of_list, LittleEndianList.le_combine_split.
compiler/src/compiler/MMIO.v:  Arguments LittleEndian.combine: simpl never. (* TODO can we put this next to its definition? *)
compiler/src/compiler/MMIO.v:  Arguments LittleEndian.split: simpl never.
compiler/src/compiler/MMIO.v:      rewrite LittleEndian.combine_split.
compiler/src/compiler/MMIO.v:      rewrite LittleEndian.combine_split.
compiler/src/compiler/MMIO.v:      rewrite LittleEndian.combine_split.
compiler/src/compiler/MMIO.v:               exists (LittleEndian.split 4 0); trivial |].
compiler/src/compiler/UniqueSepLog.v:Require Import coqutil.Word.LittleEndian.
compiler/src/compiler/UniqueSepLog.v:    bytes addr (LittleEndian.split (@Memory.bytes_per width sz) (word.unsigned value)).
compiler/src/compiler/LowerPipeline.v:        exists (List.flat_map (fun x => HList.tuple.to_list (LittleEndian.split (Z.to_nat bytes_per_word) (word.unsigned x))) stack_trash).
compiler/src/compiler/Pipeline.v:      List.flat_map (fun inst => HList.tuple.to_list (LittleEndian.split 4 (encode inst))) p.
compiler/src/compiler/FlatToRiscvDef.v:    InvalidInstruction (LittleEndian.combine 4 (HList.tuple.of_list [
compiler/src/compilerExamples/AssemblyVerif.v:    gallina_prog_1 (word.of_Z (BitOps.signExtend 32 (LittleEndian.combine 4 v1)))
compiler/src/compilerExamples/AssemblyVerif.v:                   (word.of_Z (BitOps.signExtend 32 (LittleEndian.combine 4 v2))).
compiler/src/compilerExamples/AssemblyVerif.v:  Arguments LittleEndian.combine: simpl never.
compiler/src/compilerExamples/FibCompiled.v:    rewrite LittleEndian.combine_split.
compiler/src/compilerExamples/FibCompiled.v:                  | LittleEndian.split _ _ => specialize HStore with (bs := x)
compiler/src/compilerExamples/FibCompiled.v:                  | LittleEndian.split _ _ => specialize HStore with (bs := x)
compiler/src/compilerExamples/Fibonacci.v:  List.flat_map (fun inst => HList.tuple.to_list (LittleEndian.split 4 (encode inst))) insts.
```

### DIFF
--- a/bedrock2/src/bedrock2/Memory.v
+++ b/bedrock2/src/bedrock2/Memory.v
@@ -6,7 +6,7 @@ Require Import coqutil.Decidable.
 Require Import coqutil.Datatypes.PrimitivePair coqutil.Datatypes.HList coqutil.Datatypes.List.
 Require Import coqutil.Map.Interface coqutil.Map.Properties.
 Require Import coqutil.Tactics.Tactics coqutil.Datatypes.Option.
-Require Import BinIntDef coqutil.Word.Interface coqutil.Word.LittleEndian.
+Require Import BinIntDef coqutil.Word.Interface coqutil.Word.LittleEndianList.
 Require Import bedrock2.Notations bedrock2.Syntax.
 Require Import coqutil.Byte.
 Require Import coqutil.Map.OfListWord.
@@ -141,12 +141,12 @@ Section Memory.
 
   Definition load_Z(sz: access_size)(m: mem)(a: word): option Z :=
     match load_bytes (bytes_per sz) m a with
-    | Some bs => Some (LittleEndian.combine _ bs)
+    | Some bs => Some (LittleEndianList.le_combine (tuple.to_list bs))
     | None => None
     end.
 
   Definition store_Z(sz: access_size)(m: mem)(a: word)(v: Z): option mem :=
-    store_bytes (bytes_per sz) m a (LittleEndian.split _ v).
+    store_bytes _ m a (tuple.of_list (LittleEndianList.le_split (bytes_per sz) v)).
 
   Definition load(sz: access_size)(m: mem)(a: word): option word :=
     match load_Z sz m a with
--- a/bedrock2/src/bedrock2/NotationsCustomEntry.v
+++ b/bedrock2/src/bedrock2/NotationsCustomEntry.v
@@ -128,8 +128,10 @@ Notation "output! f args" :=  (interact nil f args) (in custom bedrock_cmd at le
 Notation "unpack! lhs = f args" :=  (call lhs f args) (in custom bedrock_cmd at level 0, lhs custom bedrock_call_lhs , f custom bedrock_ident, args custom bedrock_args ).
 Notation "( lhs ) = f args" :=  (call lhs f args) (in custom bedrock_cmd at level 0, lhs custom bedrock_call_lhs , f custom bedrock_ident, args custom bedrock_args ,
   format "'(' lhs ')'  =  f args").
-Notation "f args" :=  (call nil (ident_to_string! f) args) (in custom bedrock_cmd at level 0, f ident, args custom bedrock_args ,
-  format "f args").
+Notation "f args" :=  (call nil (ident_to_string! f) args) (in custom bedrock_cmd at level 0,
+  f ident, args custom bedrock_args, format "f args").
+Notation "$ f args" :=  (call nil f args) (in custom bedrock_cmd at level 0,
+  f constr at level 0, args custom bedrock_args, format "'$' f args").
 
 Declare Scope bedrock_tail.
 Delimit Scope bedrock_tail with bedrock_tail.
@@ -309,6 +311,8 @@ Module test.
       }
     }
   )).
+
+  epose bedrock_func_body:($"write" ($1, $"W", $1, $1, $1, $1)).
 
   epose bedrock_func_body:(
   stackalloc 64 as $"b";
--- a/bedrock2/src/bedrock2/Scalars.v
+++ b/bedrock2/src/bedrock2/Scalars.v
@@ -1,4 +1,4 @@
-Require Import coqutil.Map.Interface bedrock2.Map.Separation bedrock2.Map.SeparationLogic bedrock2.Lift1Prop bedrock2.Array coqutil.Word.LittleEndian.
+Require Import coqutil.Map.Interface bedrock2.Map.Separation bedrock2.Map.SeparationLogic bedrock2.Lift1Prop bedrock2.Array coqutil.Word.LittleEndianList.
 Require Import bedrock2.Memory.
 Require Import Coq.Lists.List Coq.ZArith.ZArith.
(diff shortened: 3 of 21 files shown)


samuelgruetter:
I'm generally in favor of this change, but given that we keep claiming that our codebase is modular, I find it unsatisfactory that if Andres wants to change something in the program logic, he can only do so if Sam updates the compiler at the same time. Even though I'm planning to start using `LittleEndianList` in the compiler as well, I would like to use this opportunity to do an exercise in modularity, even if it creates some overhead compared to refactoring everything at once. I would find it reassuring to know that we can change the program logic without having to propagate changes throughout the compiler, but only having to update one proof that connects the two. And I would like you (and me as well) to do more changes to the program logic in the future, without worrying about compiler updates each time.

What do you think about this approach A:

* copy `eval_expr`, `evaluate_call_args_log` and `exec` from `bedrock2.Semantics` to `compiler.ExprImp`
* copy the original versions of `load_Z`, `store_Z`, `load` and `store` from `bedrock2.Memory` to `compiler.ExprImp`
* `compiler.ExprImp.exec` now refers to the original load/store definitions, whereas `bedrock2.Semantics.exec` refers to the new load/store definitions
* in `compiler.ExprImp`, prove `compiler.ExprImp.exec` equivalent to `bedrock2.Semantics.exec`
* in the compiler, all occurrences of `bedrock2.Semanics.exec` will have to be replaced with `compiler.ExprImp.exec`
* the program logic can now use the new load/store definitions, while the compiler can still use the old ones

On the other hand, if the work you're planning to do only depends on `WeakestPrecondition`, but not directly on `bedrock2.Semantics`, we could make the indirection a bit more lightweight, and do approach B:

* copy the original versions of `load_Z`, `store_Z`, `load` and `store` from `bedrock2.Memory` to a new module `Memory` inside file `bedrock2.Semantics`
* `bedrock2.Semantics` now refers to the original load/store definitions, whereas `bedrock2.WeakestPrecondition` refers to the new load/store definitions, so we need to fix `bedrock2.WeakestPreconditionProperties`
* in the compiler, all occurrences of `bedrock2.Memory.load/store` will have to be replaced with `bedrock2.Semantics.Memory.load/store`
* the program logic can now use the new load/store definitions, while the compiler can still use the old ones

Which approach do you prefer? Or do you have a suggestion for some other approach C?

andres-erbsen:
> the program logic can now use the new load/store definitions, while the compiler can still use the old ones

I think this part is the vast majority of work. The two tiny changes in `Memory.v` match equations I proved in coqutil, applying them in the right places really doesn't seem like it would be a lot of work. If you really want a duplicate semantics for the time being, I propose we keep the old versions in `FlatImp` and the new versions in `bedrock2.Semantics`, so the flattening proof would get the rewrites with `combine_eq` and `to_list_split`. But my perspective is that if that (or either of the alternatives you proposed) was the plan, the hard part would be getting everything else building with a tolerably small amount of duplication; the connection proof is really trivial here.

andres-erbsen:
Well, this seems to have passed with a couple of more fixes. I am thinking `Memory.WithoutTuples` or `ptsto_bytes` will be next. Do you think the kind of modularization you suggested here would be a good idea there as well?

andres-erbsen:
I'm merging this so that downstream equivalents can be merged as well. I imagine you'll find some cleanups and nits in here, but I don't think any of the changes I made are so bad as to convince us to abandon this entire direction.

samuelgruetter:
Ok, so you decided to propagate the changes through the compiler yourself because the price for modularity looked too high? :sweat_smile:  For future reference, my interpretation of this diff is that now, the interface at which we change from new (`LittleEndianList`) to old (`LittleEndian`) is the interface between the compiler and riscv-coq, more specifically the lemmas `go_fetch_inst` (for instruction memory access) and `run_load_word`/ `run_store_word` (for data memory access).
I'm still not convinced that this was the right way to do it from a modularity perspective, but if this makes the change you want in program logic compatible with the rest, why not, and we can do the modularity exercise the next time someone wants to change the program logic.

PR #229, opened 2022-02-09
Track tested branch of coqutil

JasonGross:
This will hopefully make issues like https://github.com/mit-plv/coqutil/issues/57 more rare, as dependabot updates will come from the tested branch

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,6 +4,7 @@
 [submodule "deps/coqutil"]
 	path = deps/coqutil
 	url = https://github.com/mit-plv/coqutil.git
+	branch = tested
 [submodule "deps/kami"]
 	path = deps/kami
 	url = https://github.com/mit-plv/kami.git


samuelgruetter:
What happens if I push a commit in bedrock2 that sets the coqutil submodule to a newer commit than coqutil's tested branch? Will dependabot make a PR to downgrade coqutil?

JasonGross:
I think it might make such a PR whenever the tested branch is next updated?  I'm not sure if the semantics, but we can find out;

samuelgruetter:
Sure, let's just try out! Feel free to merge this PR at whatever time you think will provide you the most insight, and if it has side effects that bother me too much, I will revert it, ok?

JasonGross:
Sounds good!

samuelgruetter:
As we saw in https://github.com/mit-plv/bedrock2/pull/308#issuecomment-1374891091, dependabot blindly makes PRs any time the submodule does not match the tip of the tracked branch, which is not what we want, because if `master` has new commits, but `tested` has not yet been fast-forwarded by the nightly script, dependabot will submit PRs that make the submodule point to the older commit. So I'm going to revert this PR.

PR #246, opened 2022-04-27
Add support for metrics in the FlatToRiscvFunctions and FlattenExpr passes

pratapsingh1729:
This PR adds support for metrics to the `FlatToRiscvFunctions` compiler pass. I changed the definition of `compiles_FlatToRiscv_correctly` to add a metrics bound requiring that the cost of executing the compiled RISC-V program is less than or equal to the cost of executing the `FlatImp` source program. I updated the proof of `compile_stmt_correct` and its various helper lemmas accordingly. 

Currently, I hardcode a fixed cost of 100 instructions, loads, and stores for each function call in `FlatImp`. This accounts for the cost of saving and restoring the caller-save registers in RISC-V (up to 29 registers in this implementation). I also added this fixed cost to the semantics of the bedrock2 source language.

~This is not yet ready to merge---some things are broken in other parts of the compiler.~

### DIFF
--- a/compiler/src/compiler/FlatToRiscvFunctions.v
+++ b/compiler/src/compiler/FlatToRiscvFunctions.v
@@ -481,34 +481,12 @@ Section Proofs.
               (of_list
                  (list_union Z.eqb (List.firstn binds_count (reg_class.all reg_class.arg)) []))
               (singleton_set RegisterNames.ra)) (getRegs finalL) /\
-
-(* (getMetrics finalL - *)
-(*   (Platform.MetricLogging.addMetricInstructions 1 *)
-(*      (Platform.MetricLogging.addMetricJumps 1 *)
-(*         (Platform.MetricLogging.addMetricLoads 1 *)
-(*            (Platform.MetricLogging.addMetricInstructions 1 *)
-(*               (Platform.MetricLogging.addMetricLoads 1 *)
-(*                  (Platform.MetricLogging.addMetricInstructions 1 *)
-(*                     (Platform.MetricLogging.addMetricLoads 2 *)
-(*                        (Platform.MetricLogging.addMetricInstructions *)
-(*                           #(Datatypes.length *)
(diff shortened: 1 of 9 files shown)


pratapsingh1729:
The `FlattenExpr` pass also compiles now---I added the fixed cost for saving and restoring caller-save registers to the bedrock2 source semantics.

This now compiles without errors on my machine---it should be mergeable I think.

samuelgruetter:
closing and reopening to trigger a build

compiler/src/compiler/FlatToRiscvFunctions.v
```diff
@@ -1032,9 +1046,61 @@ Section Proofs.
...
+             | |- context[?f ?n (Platform.MetricLogging.mkMetricLog ?i ?s ?l ?j)] =>
+                 match type of (f n (Platform.MetricLogging.mkMetricLog i s l j)) with
+              | Platform.MetricLogging.MetricLog =>
+                  let t := eval hnf in (f n (Platform.MetricLogging.mkMetricLog i s l j)) in
+                    let t' := eval cbn in t in
+                      change (f n (Platform.MetricLogging.mkMetricLog i s l j)) with t' in H
+                 end
+             end.
+      cbn in H2p6.
+      blia.
```
samuelgruetter, line 1098: Could this code replace the metric solving tactic in general?

compiler/src/compiler/FlatToRiscvFunctions.v
```diff
@@ -1528,33 +1596,30 @@ Section Proofs.
...
+      1:{ 
+              match goal with
+              | H: (binds_count <= 8)%nat |- _ => rename H into BC
+              end.
+              move BC after OC.
+              repeat match goal with
+                     | x := _ |- _ => clearbody x
+                     end.
+              
+              Search initialL_metrics.
```
samuelgruetter, line 1609: `Search` commands should not be left in the code

PR #265, opened 2022-07-21
Fix probable typo introduced in bcd4bcf014b54db0137a534472baa32c9fd35200

JasonGross:


andres-erbsen:
I agree it looks like a typo; would have been caught by a single test (or proof against https://github.com/mit-plv/fiat-crypto/blob/bf49a3f2bb9d827bcb829e946d50bee415e1444f/src/Bedrock/End2End/RupicolaCrypto/Spec.v#L55= )

PR #283, opened 2022-10-06
Pass `-native-compiler ondemand` for bedrock2 ex

### LINKED ISSUE #282: bedrock2 lightbulb example is incompatible with native_compute
Fiat Crypto is failing on Coq's CI with
```
COQNATIVE /builds/coq/coq/_build_ci/fiat_crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/lightbulb.vo
ocamlopt.opt got signal and exited
Error: Native compiler exited with status 2 (in case of stack overflow,
       increasing stack size (typically with "ulimit -s") often helps)
Command exited with non-zero status 1
/builds/coq/coq/_build_ci/fiat_crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/lightbulb.vo.native (real: 5214.51, user: 60.09, sys: 13.67, mem: 3542188 ko)
```
https://gitlab.com/coq/coq/-/jobs/3117535414
https://coq.zulipchat.com/#narrow/stream/237656-Coq-devs-.26-plugin-devs/topic/fiat.20crypto.20broken

Are there any `Defined` that can be replaced with `Qed`?  Or is there a way to make Fiat Crypto depend only on a subset of bedrock2_ex that does not include lightbulb?  Or can bedrock2 be configured to pass `-native-compiler ondemand` or whatever when building examples?

JasonGross:
Fixes #282

### DIFF
--- a/bedrock2/Makefile
+++ b/bedrock2/Makefile
@@ -56,7 +56,7 @@ Makefile.coq.noex: force _CoqProject
 
 Makefile.coq.ex: force _CoqProject
 	@echo "Generating Makefile.coq.ex"
-	@$(COQ_MAKEFILE) $(VS_EX) -o Makefile.coq.ex
+	@$(COQ_MAKEFILE) $(VS_EX) -arg -native-compiler -arg ondemand -o Makefile.coq.ex
 
 BYTEDUMP = COQFLAGS="$(ALLDEPFLAGS)" ../etc/bytedump.sh
 


JasonGross:
The Coq packages we use on CI are configured in such a way that coqnative is not run by default on builds.  Note that we do this already in Fiat Crypto, I believe, which has similar problems in some places.  Plausibly there's some large enough amount of stack and RAM that would allow coqnative to work here, but I don't think it's worth it, and I don't expect many people to suffer for not having the definitions from bedrock2 examples precompiled to native code

andres-erbsen:
But do Coq developers think that the supported usage of Coq is to automatically native-compile everything that ends in Defined?

PR #292, opened 2022-10-25
Paranoid to c string

andres-erbsen:


bedrock2/src/bedrock2/ToCString.v
```diff
@@ -40,6 +38,34 @@ static __attribute__((always_inline)) inline void
 _br2_store(uintptr_t a, uintptr_t v, uintptr_t sz) {
   memcpy((void*)a, &v, sz);
 }
+
+static __attribute__((always_inline)) inline uintptr_t
+_br2_mulhuu(uintptr_t a, uintptr_t b) {
+#if (UINTPTR_MAX == (1LLU<<31) - 1 + (1LLU<<31))
+	return ((uint64_t)a * b) >> 32;
+#elif (UINTPTR_MAX == (1LLU<<63) - 1 + (1LLU<<63))
+	return ((__uint128_t)a * b) >> 64;
```
JasonGross, line 47: Is there a reason to use this undocumented type rather than
```suggestion
	return ((unsigned __int128)a * b) >> 64;
```
which is documented?

PR #313, opened 2023-01-13
Initial interpreter for expressions

hulsemohit:


PyLevelLang/src/PyLevelLang/Interpret.v
```diff
@@ -1,5 +1,8 @@
 Require Import PyLevelLang.Language.
 Require Import coqutil.Map.Interface.
+Require Import coqutil.Map.Interface coqutil.Map.SortedListString.
```
samuelgruetter, line 3:
```suggestion
Require Import coqutil.Map.Interface coqutil.Map.SortedListString.
```

PyLevelLang/src/PyLevelLang/Interpret.v
```diff
@@ -13,14 +16,118 @@ Fixpoint interp_type (t : type) :=
...
+                    | OLengthString => fun e1 => Z.of_nat (String.length (interp_expr l e1))
+                    | OFst _ _ => fun e1 => fst (interp_expr l e1)
+                    | OSnd _ _ => fun e1 => snd (interp_expr l e1)
+                    end e1
+    | EBinop o e1 e2 => match o in binop t1 t2 t3 return expr t1 -> expr t2 -> interp_type t3 with
+                        | OPlus => fun e1 e2 => interp_expr l e1 + interp_expr l e2
+                        | OMinus => fun e1 e2 => interp_expr l e1 - interp_expr l e2
+                        | OTimes => fun e1 e2 => interp_expr l e1 * interp_expr l e2
+                        | ODiv => fun e1 e2 => interp_expr l e1 / interp_expr l e2
+                        | OMod => fun e1 e2 => Zmod (interp_expr l e1) (interp_expr l e2)
```
samuelgruetter, line 80: `Zmod` is legacy, I'd use `Z.mod` or infix `mod`

PyLevelLang/src/PyLevelLang/Interpret.v
```diff
@@ -13,14 +16,118 @@ Fixpoint interp_type (t : type) :=
...
+                        | OLess => fun e1 e2 => (interp_expr l e1) <? (interp_expr l e2)
+                        | OEq _ => fun e1 e2 => compare_values (interp_expr l e1) (interp_expr l e2)
+                        | ORepeat _ => fun e1 e2 => repeat (interp_expr l e2) (Z.to_nat (interp_expr l e1))
+                        | OPair _ _ => fun e1 e2 => (interp_expr l e1, interp_expr l e2)
+                        | OCons _ => fun e1 e2 => interp_expr l e1 :: interp_expr l e2
+                        | ORange => fun e1 e2 => 
+                            let lo := interp_expr l e1 in 
+                            let hi := interp_expr l e2 in 
+                            eval_range lo (Z.to_nat (hi - lo))
+                        end e1 e2
```
samuelgruetter, line 94: Here you're repeating 15x the fact that in order to evaluate a binary operator whose arguments are `e1` and `e2`, you have to compute `interp_expr l e1` and `interp_expr l e2`. You could deduplicate this with the following helper function:

```coq
Definition interp_binop {t1 t2 t3 : type} (o : binop t1 t2 t3) :
  interp_type t1 -> interp_type t2 -> interp_type t3 :=
  match o in binop t1 t2 t3 return interp_type t1 -> interp_type t2 -> interp_type t3 with
  | OPlus => Z.add
  | OMinus => Z.sub
  | OTimes => Z.mul
   ...
  end.
```

samuelgruetter: And while we're at it, I'd also recommend to factor out the cases for `EVar`, `ELoc`, `EConst` and `EUnop` into separate helper functions, because `interp_expr` will appear in many lemma statements and proof states, so having small functions will make everything more concise and readable.

PyLevelLang/src/PyLevelLang/Interpret.v
```diff
@@ -13,14 +16,118 @@ Fixpoint interp_type (t : type) :=
...
+  Goal interp_expr map.empty ex2 = 42.
+  reflexivity. Qed.
+
+  Local Open Scope string_scope.
+  Definition ex3 : expr (TPair TInt (TPair TBool TString)) :=
+      (EBinop (OPair _ _) (EConst (CInt 42))
+        (EBinop (OPair _ _) (EConst (CBool true)) (EConst (CString "hello")))).
+  Goal interp_expr map.empty ex3 = (42, (true, "hello")).
+  reflexivity. Qed.
+End Examples.
```
samuelgruetter, line 133: It would be cool to move the examples for the elaborator and for the interpreter into a separate test file that imports both the elaborator and the interpreter and composes the two. And soon, also composes with Leo's notations.

samuelgruetter (commented):
Looks great! Some comments below:

hulsemohit:
Thanks for the feeback!I've refactored most of the `interp_expr` into smaller functions.

PyLevelLang/src/PyLevelLang/Examples.v
```diff
@@ -0,0 +1,79 @@
...
+  Instance tenv_ok : map.ok tenv := SortedListString.ok _.
+  
+  Instance locals : map.map string {t & interp_type t} := SortedListString.map _.
+  Instance locals_ok : map.ok locals := SortedListString.ok _.
+
+  Definition elaborate_interpret (p : pexpr) : result {t & interp_type t} :=
+    match elaborate map.empty p with
+    | Success v => Success (existT _ _ (interp_expr map.empty (projT2 v)))
+    | Failure e => Failure e
+    end.
```
samuelgruetter, line 19: This could be written using `_ <- _;; _` notation, and could also take a `locals` argument (which you'd have to transform into a `tenv` to pass to `elaborate` using `map.map_values`, which you can get with `Require Import coqutil.Map.Properties`)

PyLevelLang/src/PyLevelLang/Interpret.v
```diff
@@ -13,14 +15,101 @@ Fixpoint interp_type (t : type) :=
...
+  | left H => cast H _ (projT2 v)
+  | _ => default_val t_expected
+  end.
+
+Definition eqb_values {t : type} : interp_type t -> interp_type t -> bool :=
+  match t with
+  | TInt => Z.eqb
+  | TString => String.eqb
+  | TBool => Bool.eqb
+  | _ => fun _ _ => false
```
samuelgruetter, line 44: Let's not forget that if we don't want to support equality of non-primitive values, we need to update the elaborator to reject these

PyLevelLang/src/PyLevelLang/Interpret.v
```diff
@@ -13,14 +15,101 @@ Fixpoint interp_type (t : type) :=
...
+    | EVar _ x => get_local l x
+    | ELoc _ x => get_local l x
+    | EConst c => interp_const c
+    | EUnop o e1 => interp_unop l o (interp_expr l e1)
+    | EBinop o e1 e2 => interp_binop l o (interp_expr l e1) (interp_expr l e2)
+    | EFlatmap e1 x e2 => flat_map (fun y => interp_expr (set_local l x y) e1) (interp_expr l e2)
+    | EIf e1 e2 e3 => match interp_expr l e1 with
+                      | true => interp_expr l e2
+                      | false => interp_expr l e3
+                      end
```
samuelgruetter, line 111: this could be an if-then-else

samuelgruetter (commented):
Neat! And yes, a non-empty `locals` for testing would be good, see comment below

PyLevelLang/src/PyLevelLang/Examples.v
hulsemohit: I added a `locals` argument, but this way of implementing it has an issue - the `locals` type contains no information about whether a binding is a variable or a location, so I need to default to one kind in for `tenv` (I've defaulted to locations).

hulsemohit: One way to fix this might be to also take `tenv` as an argument and check compatibility with `locals` (and failing if they aren't).

PyLevelLang/src/PyLevelLang/Interpret.v
hulsemohit: I added a new function `can_eq`, which takes a `type` and checks if equality is supported. The elaborator now enforces this.
I also added support for equating `TEmpty` (always true, since there's only one value).

PyLevelLang/src/PyLevelLang/Examples.v
samuelgruetter: I think defaulting to locations is fine! That just means that we run our tests in some environment of mutable locals, which seems fine.

samuelgruetter:
Looks great, merging now!

PR #314, opened 2023-01-15
[do not merge] work-in-progress UseImmediate proof

0adb:
Work-in-progress for proposed phase of compilation between bedrock2 and RISC, as future compiler optimization (using RISC-V immediate instructions).


compiler/src/compiler/UseImmediateDef.v
```diff
@@ -0,0 +1,132 @@
...
+                                    else default
+            | Syntax.bopname.sub => if (is12BitImmediate (-l1)) then
+                                      if eqb v1 v2b then
+                                        SOp v2 Syntax.bopname.add v2a (Const (-l1))
+                                      else default
+                                    else default
+            | _ => default
+            end
+        | SLit v1 l1, SSeq (SLit v1b l1b) (SOp v2 op v2a (Var v2b)) =>
+            if (eqb v1 v1b) then default
```
samuelgruetter, line 65: Why is this case needed?

compiler/src/compiler/UseImmediateDef.v
```diff
@@ -0,0 +1,132 @@
...
+    | SIf c s1 s2 => SIf c (useImmediate s1) (useImmediate s2)
+    | SLoop s1 c s2 => SLoop (useImmediate s1) c (useImmediate s2)
+    | SStackalloc v1 sz1 s => SStackalloc v1 sz1 (useImmediate s)
+    | SSeq s1 s2 =>
+        let used1 := useImmediate s1 in
+        let used2 := useImmediate s2 in
+        let default := SSeq used1 used2 in
+        match used1, used2 with
+        | SSkip, _ => used2
+        | _, SSkip => used1
```
samuelgruetter, line 31: These two cases are unnecessary: It doesn't hurt to have a superfluous `SSkip`, because they will be translated into the empty list of instructions.

compiler/src/compiler/UseImmediateDef.v
0adb: I'm not sure this case would come up realistically speaking in terms of the inputs that UseImmediate would get, but in case flattened code somehow came out to
```
t1 = 3
t1 = 2**25
t2 = t1 + t1
```
the second argument `SSeq (SLit v1b l1b) (SOp v2 op v2a (Var v2b))` would exist because 2**25 is too big to fit into assembly, but if the eqb v1 v1b check isn't there, trying to propagate constant 3 would be bad because it would turn it into
```
t1 = 3
t1 = 2**25
t2 = t1 + 3
```
which isn't correct. However, if instead the flattened code was 
```
t1 = 3
t2 = 2**25
t3 = t1 + t2
```
then since eqb v1 v1b is false, it's fine to try to convert the code into
```
t1 = 3
t2 = 2**25
t3 = t2 + 3
```.
At least this is what I think is why this case should be there but not completely sure.

File: compiler/src/compiler/UseImmediateDef.v (hunk 2 above)
Comment (0adb) [reply compiler/src/compiler/UseImmediateDef.v:31]: I think I included this because I didn't want to include a case for `SLit v1 l1, SSeq (SSkip) (SOp v2 op v2a (Var v2b))` which would have the same cases and expressions as `SLit v1 l1, SOp v2 op v2a (Var v2b)`, but also in hindsight I think I wouldn't have to duplicate code and instead could do 
```
| SLit v1 l1, SSeq (SSkip) (SOp v2 op v2a (Var v2b))
| SLit v1 l1, SOp v2 op v2a (Var v2b) => (* insert same code here *)
```
.
I'm unsure about which of removing the `SSkip, _` cases and adding
```
| SLit v1 l1, SSeq (SSkip) (SOp v2 op v2a (Var v2b))
| SLit v1 l1, SOp v2 op v2a (Var v2b) => (* insert same code here *)
```
should be done here.

File: compiler/src/compiler/UseImmediateDef.v (hunk 1 above)
Comment (samuelgruetter) [reply compiler/src/compiler/UseImmediateDef.v:65]: I see, so if you want to treat the case ` | SLit v1 l1, SSeq (SLit v1b l1b) (SOp v2 op v2a (Var v2b)) =>`, then it makes sense that you also need an `if (eqb v1 v1b) then default`. But is it really worth treating the case ` | SLit v1 l1, SSeq (SLit v1b l1b) (SOp v2 op v2a (Var v2b)) =>` specially? Also keep in mind that you don't have to implement constant propagation here: If a program contains something like
```
t1 = some constant;
t2 = some other constant;
some code not accessing t1 nor t2;
t3 = t1 + t2
```
a separate constant propagation phase can take care of replacing this with
```
t3 = result of adding some constant and some other constant
```
(if `t1` and `t2` are not read after that snippet).

Or asked differently: Can you write a source program (like `ex0`, `ex1` etc below) for which `useImmediate` only does the right thing if you keep the case ` | SLit v1 l1, SSeq (SLit v1b l1b) (SOp v2 op v2a (Var v2b)) =>`, and doesn't do what you want it to do if you remove that case?

File: compiler/src/compiler/UseImmediateDef.v [hunk 3]
```diff
@@ -0,0 +1,132 @@
...
+  := match t with
+     | (targs, trets, tstmt) => (targs, trets, (useImmediate (fun x => true) (fun x => Z.ltb x 16) tstmt))
+     end.
+
+Definition immediateExs'' := map useImmOnFlats'' unwrappedFlattenExs.
+
+Compute unwrappedFlattenExs.
+Compute immediateExs. (* All constants are small enough. *)
+Compute immediateExs'. (* is5BitImmediate is true only for < 16. *)
+Compute immediateExs''. (* is12BitImmediate is true only for < 16. *)
```
samuelgruetter, line 132: Doing some tests on the ouput of the flattening phase like you did here is great :+1: 
To make the code more robust and readable, and avoid that output of `Compute` gets printed to stdout on each build, you could do the something like the following for each example:
```coq
Goal useImmediate (flatten_function ex_i) = expected_output. vm_compute. reflexivity. Abort.
```
Like this, each time this file is compiled, Coq also checks that `useImmediate` outputs the expected results on the examples, so if you modify `useImmediate` in a way that breaks the examples, you'll notice immediately without having to look at each output.

compiler/src/compiler/UseImmediateDef.v
samuelgruetter: You don't need a `| SLit v1 l1, SSeq (SSkip) (SOp v2 op v2a (Var v2b))` case, because if it turns out that there are such superfluous `SSkip` statements that prevent your `useImmediate` from kicking in, we can simply add a separate phase that first removes all superfluous `SSkip` commands.

compiler/src/compiler/UseImmediate.v
```diff
@@ -0,0 +1,33 @@
...
+    intros.
+    induction H; simpl; eauto.
+
+    destruct (useImmediate is5BitImmediate is12BitImmediate s1) eqn:Es1; destruct (useImmediate is5BitImmediate is12BitImmediate s2) eqn:Es2; try eauto.
+
+    destruct z eqn:Ez; destruct op eqn:Eop; try eauto.
+    { destruct (is12BitImmediate v) eqn:Ev.
+      { destruct (x =? v0)%string eqn:Exv0.
+        { eapply exec.op.
+          2: { simpl. reflexivity. }
```
0adb, line 32: Currently the proof goal at this point is `map.get l y = Some ?y'` but using `assumption` or `eassumption` doesn't work. (The detailed output is below.) I tried using `destruct (map.get l y) eqn:Eg; ` on line 28 before the other `destruct`'s to do casework on whether `y` is defined in locals `l` but I'm not sure if this is right. The output with using `destruct (map.get l y)`  under case `map.get l y = Some r` is the second proof output below.
```

  width : Z
  BW : Bitwidth.Bitwidth width
  word : Interface.word width
  env : map.map string (list string * list string * stmt string)
  mem : map.map word (Init.Byte.byte : Type)
  locals : map.map string word
  ext_spec : Semantics.ExtSpec
  is5BitImmediate, is12BitImmediate : Z -> bool
  e : env
  t : Semantics.trace
  m : mem
  l : locals
  mc : MetricLogging.MetricLog
  s1, s2 : stmt string
  mid, post : Semantics.trace -> mem -> locals -> MetricLogging.MetricLog -> Prop
  H : exec e s1 t m l mc mid
  H0 : forall (t' : Semantics.trace) (m' : mem) (l' : locals)
         (mc' : MetricLogging.MetricLog), mid t' m' l' mc' -> exec e s2 t' m' l' mc' post
  x0 : string
  op : bopname
  y : string
  z : operand
  v0 : string
  Ez : z = Var v0
  Eop : op = bopname.add
  Es2 : useImmediate is5BitImmediate is12BitImmediate s2 = SOp x0 bopname.add y (Var v0)
  H1 : forall (t' : Semantics.trace) (m' : mem) (l' : locals)
         (mc' : MetricLogging.MetricLog),
       mid t' m' l' mc' -> exec e (SOp x0 bopname.add y (Var v0)) t' m' l' mc' post
  x : string
  v : Z
  Es1 : useImmediate is5BitImmediate is12BitImmediate s1 = SLit x v
  IHexec : exec e (SLit x v) t m l mc mid
  Ev : is12BitImmediate v = true
  Exv0 : (x =? v0)%string = true
  ============================
  map.get l y = Some ?y'

goal 2 (ID 5131) is:
 post t m (map.put l x0 (Semantics.interp_binop bopname.add ?y' (word.of_Z v)))
   (MetricLogging.addMetricLoads 2 (MetricLogging.addMetricInstructions 2 mc))
```
```
  width : Z
  BW : Bitwidth.Bitwidth width
  word : Interface.word width
  env : map.map string (list string * list string * stmt string)
  mem : map.map word (Init.Byte.byte : Type)
  locals : map.map string word
  ext_spec : Semantics.ExtSpec
  is5BitImmediate, is12BitImmediate : Z -> bool
  e : env
  t : Semantics.trace
  m : mem
  l : locals
  mc : MetricLogging.MetricLog
  s1, s2 : stmt string
  mid, post : Semantics.trace -> mem -> locals -> MetricLogging.MetricLog -> Prop
  H : exec e s1 t m l mc mid
  H0 : forall (t' : Semantics.trace) (m' : mem) (l' : locals)
         (mc' : MetricLogging.MetricLog),
       mid t' m' l' mc' -> exec e s2 t' m' l' mc' post
  x0 : string
  op : bopname
  y : string
  z : operand
  v0 : string
  Ez : z = Var v0
  Eop : op = bopname.add
  Es2 : useImmediate is5BitImmediate is12BitImmediate s2 =
        SOp x0 bopname.add y (Var v0)
  H1 : forall (t' : Semantics.trace) (m' : mem) (l' : locals)
         (mc' : MetricLogging.MetricLog),
       mid t' m' l' mc' ->
       exec e (SOp x0 bopname.add y (Var v0)) t' m' l' mc' post
  x : string
  v : Z
  Es1 : useImmediate is5BitImmediate is12BitImmediate s1 = SLit x v
  IHexec : exec e (SLit x v) t m l mc mid
  r : word
  Eg : map.get l y = Some r
  Ev : is12BitImmediate v = true
  Exv0 : (x =? v0)%string = true
  ============================
  post t m (map.put l x0 (word.add r ?z'))
    (MetricLogging.addMetricLoads 2 (MetricLogging.addMetricInstructions 2 mc))
```

samuelgruetter: How about this: Don't `eapply exec.op`, and early enough, do `inversion IHexec`, which will tell you that `mid` holds for `l` with `x` updated to `v`. Then you can use this to specialize `H1` (which, btw, is also an IH, but just not named correctly by the induction tactic).
Also, if you use `destr` instead of `destruct`, it will automatically add an equation, and also prettify that equation, eg instead of leaving you with a `(x =? v0)%string = true`, it will derive `x = v0` and (I believe) also substitute `x`.

compiler/src/compiler/UseImmediate.v
```diff
@@ -0,0 +1,32 @@
...
+    forall e st t m l mc post, exec e st t m l mc post -> exec e (useImmediate is5BitImmediate is12BitImmediate st) t m l mc post.
+  Proof.
+    intros.
+    induction H; simpl; eauto.
+
+    destr (useImmediate is5BitImmediate is12BitImmediate s1); destr (useImmediate is5BitImmediate is12BitImmediate s2); try eauto.
+    destr (map.get l y); destr z; destr op; progress inversion IHexec.
+    { destr (is12BitImmediate v).
+      { destr (x =? v0)%string.
+        { specialize H1 with (t' :=  t) (m' := m) (l' := (map.put l v0 (word.of_Z v))) (mc' := (MetricLogging.addMetricLoads 8 (MetricLogging.addMetricInstructions 8 mc))). clear H2 H3 H4 H5 H6 H7 H8. apply H0 in H9 as H2. apply H1 in H9 as H3.
```
0adb, line 30: Would like to ask if this was the general direction that was being hinted at from earlier comments. Currently the proof goal looks like 
```
  
  v0 : string
  width : Z
  BW : Bitwidth.Bitwidth width
  word : Interface.word width
  env : map.map string (list string * list string * stmt string)
  mem : map.map word (Init.Byte.byte : Type)
  locals : map.map string word
  ext_spec : Semantics.ExtSpec
  is5BitImmediate, is12BitImmediate : Z -> bool
  e : env
  t : Semantics.trace
  m : mem
  l : locals
  mc : MetricLogging.MetricLog
  s1, s2 : stmt string
  mid, post : Semantics.trace -> mem -> locals -> MetricLogging.MetricLog -> Prop
  H : exec e s1 t m l mc mid
  H0 : forall (t' : Semantics.trace) (m' : mem) (l' : locals)
         (mc' : MetricLogging.MetricLog),
       mid t' m' l' mc' -> exec e s2 t' m' l' mc' post
  x0, y : string
  E0 : useImmediate is5BitImmediate is12BitImmediate s2 =
       SOp x0 bopname.add y (Var v0)
  v : Z
  H1 : mid t m (map.put l v0 (word.of_Z v))
         (MetricLogging.addMetricLoads 8
            (MetricLogging.addMetricInstructions 8 mc)) ->
       exec e (SOp x0 bopname.add y (Var v0)) t m (map.put l v0 (word.of_Z v))
         (MetricLogging.addMetricLoads 8
            (MetricLogging.addMetricInstructions 8 mc)) post
  IHexec : exec e (SLit v0 v) t m l mc mid
  E : useImmediate is5BitImmediate is12BitImmediate s1 = SLit v0 v
  r : word
  E1 : map.get l y = Some r
  t0 : Semantics.trace
  m0 : mem
  l0 : locals
  mc0 : MetricLogging.MetricLog
  x1 : string
  v1 : Z
  post0 : Semantics.trace -> mem -> locals -> MetricLogging.MetricLog -> Prop
  H9 : mid t m (map.put l v0 (word.of_Z v))
         (MetricLogging.addMetricLoads 8
            (MetricLogging.addMetricInstructions 8 mc))
  E2 : is12BitImmediate v = true
  H2 : exec e s2 t m (map.put l v0 (word.of_Z v))
         (MetricLogging.addMetricLoads 8
            (MetricLogging.addMetricInstructions 8 mc)) post
  H3 : exec e (SOp x0 bopname.add y (Var v0)) t m (map.put l v0 (word.of_Z v))
         (MetricLogging.addMetricLoads 8
            (MetricLogging.addMetricInstructions 8 mc)) post
  ============================
  exec e (SOp x0 bopname.add y (Const v)) t m l mc post
```

compiler/src/compiler/UseImmediateDef.v
```diff
@@ -0,0 +1,138 @@
...
+  Fixpoint useImmediate(s: stmt string) : stmt string :=
+    match s with
+    | SIf c s1 s2 => SIf c (useImmediate s1) (useImmediate s2)
+    | SLoop s1 c s2 => SLoop (useImmediate s1) c (useImmediate s2)
+    | SStackalloc v1 sz1 s => SStackalloc v1 sz1 (useImmediate s)
+    | SSeq s1 s2 =>
+        let used1 := useImmediate s1 in
+        let used2 := useImmediate s2 in
+        let default := SSeq used1 used2 in
+        match used1, used2 with
```
samuelgruetter, line 30: Why do you match here on `used1, used2` instead of on `s1, s2`? Does that make the transformation more powerful or more correct? From looking at your current proof goal, it seems to me that matching on `s1, s2` would simplify the proof.

compiler/src/compiler/UseImmediateDef.v
```diff
@@ -0,0 +1,139 @@
...
+        | _, SSkip => used1 *)
+        | SLit v1 l1, SOp v2 op v2a (Var v2b) =>
+
+            match op with
+            | Syntax.bopname.add
+            | Syntax.bopname.and
+            | Syntax.bopname.or
+            | Syntax.bopname.xor => if (is12BitImmediate l1) then
+                                      if eqb v1 v2b then
+                                        SSeq s1 (SOp v2 op v2a (Const l1))
```
0adb, line 40: Would like to ask if it's right to have replaced `(SOp v2 op v2a (Const l1))` with `SSeq s1 (SOp v2 op v2a (Const l1))`. I made change because I realized that originally the change in locals that `s1` does is lost if we don't keep `s1` in the output but not sure if this is right.

samuelgruetter: Yes, that looks correct! (And we'll run a separate dead-code elimination phase later to get rid of the `SLit` commands that became unnecessary)

compiler/src/compiler/UseImmediate.v
```diff
@@ -0,0 +1,46 @@
...
+             { inversion H. clear H1 H2 H3 H4 H5 H6 H7.  specialize H0 with (t' := t) (m' := m) (l' := (map.put l v0 (word.of_Z v))) (mc' := (MetricLogging.addMetricLoads 8 (MetricLogging.addMetricInstructions 8 mc))). apply H0 in H8. clear H0. apply exec.seq_cps.  apply exec.lit. inversion H8. clear H0 H1 H2 H3 H4 H5 H6 H7 H9. simpl in H11. rewrite  map.get_put_same in H11. fwd. eapply exec.op.
+               { apply H10. }
+               { simpl. reflexivity. }
+               { apply H12. }
+             }
+             { destr ((x =? y)%string).
+               { inversion H.  clear H1 H3 H2 H4 H5 H6 H7.  specialize H0 with (t' := t) (m' := m) (l' := (map.put l y (word.of_Z v))) (mc' := (MetricLogging.addMetricLoads 8 (MetricLogging.addMetricInstructions 8 mc))). apply H0 in H8.  clear H0. apply exec.seq_cps. apply exec.lit. inversion H8. clear H0 H1 H2 H3 H4 H5 H6 H7 H9. simpl in H11. rewrite map.get_put_same in H10. fwd. eapply exec.op.
+                 { apply H11. }
+                 { simpl. reflexivity. }
+                 { idtac. progress simpl in *.  Search word.add. apply H12. }
```
0adb, line 39: Currently I'm at this part of the proof, where I have a hypothesis
```
H12 : post t m
          (map.put (map.put l y (word.of_Z v)) x0
             (word.add (word.of_Z v) z'))
          (MetricLogging.addMetricLoads 2
             (MetricLogging.addMetricInstructions 2
                (MetricLogging.addMetricLoads 8
                   (MetricLogging.addMetricInstructions 8 mc))))
```
and a proof goal
```
post t m
    (map.put (map.put l y (word.of_Z v)) x0
       (word.add z' (word.of_Z v)))
    (MetricLogging.addMetricLoads 2
       (MetricLogging.addMetricInstructions 2
          (MetricLogging.addMetricLoads 8
             (MetricLogging.addMetricInstructions 8 mc)))
```
which are almost the same except for the order of arguments to `word.add` being switched in order, and it seems that there should be a commutativity of addition proposition to use but searching word.add doesn't seem to show it.

compiler/src/compiler/UseImmediate.v
```diff
@@ -33,34 +33,131 @@ Section WithArguments.
...
+               { simpl. reflexivity. }
+               { apply H12. }
+             }
+             { destr ((x =? y)%string).
+               { inversion H.  clear H1 H3 H2 H4 H5 H6 H7.  eapply H0 in H8.  clear H0. apply exec.seq_cps. apply exec.lit. inversion H8. clear H0 H1 H2 H3 H4 H5 H6 H7 H9. simpl in H11. rewrite map.get_put_same in H10. fwd. eapply exec.op.
+                 { apply H11. }
+                 { simpl. reflexivity. }
+                 { progress simpl in *. replace (word.and z' (word.of_Z v))  with  (word.and (word.of_Z v) z').
+                   { apply H12. }
+                   { Search word.and. induction (word.of_Z v).
```
0adb, line 122: I'm currently stuck on goal `word.and (word.of_Z v) z' = word.and z' (word.of_Z v)`, and unsure how to progress. Things that have failed include:

- Using the `ring` tactic.
- Based on proof of `unsigned_and_wrap` in `coqutil/Word/Properties.v`, tried to copy over `Hint Rewrite
       unsigned_of_Z signed_of_Z of_Z_unsigned unsigned_add unsigned_sub unsigned_opp unsigned_or unsigned_and unsigned_xor unsigned_not unsigned_ndn unsigned_mul signed_mulhss signed_mulhsu unsigned_mulhuu unsigned_divu signed_divs unsigned_modu signed_mods unsigned_slu unsigned_sru signed_srs unsigned_eqb unsigned_ltu signed_lts using trivial
  : word_laws.` followed by 
  `Ltac bitwise :=
      autorewrite with word_laws;
      unfold wrap;
      generalize_wrap_unsigned;
      Z.bitblast.` , but got error that `unsigned_of_Z` wasn't a recognized symbol.
- Using `Search word.and` also seems to turn up `Word.build_ok` which has something like `word.unsigned (word.and x y) =  wrap (Z.land (word.unsigned x) (word.unsigned y)))` but mixed with a lot of other propositions in an implies chain.
- Also tried `induction (word.of_Z v)` and got error "Unable to satisfy the following constraints: ?word : "Interface.word ?width"

samuelgruetter: Yes, our word `ring` doesn't work here because it's only for arithmetic, but `word.and` is a bitwise operation.

So for this kind of goals, you need to be aware of our *word goal recipe*: :wink:

* First, `eapply word.unsigned_inj` to turn the word goal into a Z goal.
* Now you have an equality between two `word.unsigned (word.OP args)`. Find lemmas that allow you to rewrite `word.unsigned (word.OP arg ...)` into `Z.OP (word.unsigned arg) ...`. They are typically named `word.unsigned_OP`, eg `word.unsigned_and`.
* Now that you have operations on Z instead of word, you can use all the standard library lemmas about Z. So in this case, you could do `Search Z.land "comm"` or `Search (Z.land ?x ?y = Z.land ?y ?x)`, which returns you exactly the lemma you need. Rewrite with it, and `reflexivity`, done!

There might be several lemmas of this kind that would be useful to have, but are currently missing (`word.add_comm`, `word.xor_comm`, ...). I would first prove them right in the file you're working on to avoid too long recompilation times, but once you have all that you need, it would be good to put them into `coqutil.Word.Properties`.

compiler/src/compiler/UseImmediate.v
```diff
@@ -0,0 +1,355 @@
...
+  Local Hint Constructors exec: core.
+
+  Add Ring wring : (word.ring_theory (word := word))
+      (preprocess [autorewrite with rew_word_morphism],
+       morphism (word.ring_morph (word := word)),
+        constants [word_cst]).
+
+  Lemma word_and_comm:
+    forall {width: Z} {word: Interface.word width},
+      word.ok word -> forall x y : word, word.and x y = word.and y x.
```
samuelgruetter, line 28: If you already quantify over `{width : Z}`,  `{word :  word width}` and `{word_ok : word.ok word}` using the `Context` command above, you don't need to list them here again, because each definition and lemma inside the section will automatically quantify over them. Once you `End` the section, you can use `Check word_and_comm.` to see how the section arguments were added.
And, actually, it would also be good to move them to `coqutil.Word.Properties`.

compiler/src/compiler/UseImmediate.v
```diff
@@ -0,0 +1,355 @@
...
+            { apply H12. }
+          }
+       }
+       {
+         inversion H. clear H1 H2 H3 H4 H5 H6 H7. eapply H0 in H8. clear H0. apply exec.seq_cps. apply exec.lit. inversion H8. clear H0 H1 H2 H3 H4 H5 H6 H7 H9. simpl in H11. fwd. eapply exec.op.
+         { apply H10. }
+         { simpl. reflexivity. }
+         { apply H12. }
+       }
+Qed.
```
samuelgruetter, line 353: Congrats on the `Qed`, good job :clap: :tada:

compiler/src/compiler/UseImmediate.v
```diff
@@ -0,0 +1,355 @@
...
+  Add Ring wring : (word.ring_theory (word := word))
+      (preprocess [autorewrite with rew_word_morphism],
+       morphism (word.ring_morph (word := word)),
+        constants [word_cst]).
+
+  Lemma word_and_comm:
+    forall {width: Z} {word: Interface.word width},
+      word.ok word -> forall x y : word, word.and x y = word.and y x.
+  Proof.
+    intros. eapply word.unsigned_inj. rewrite word.unsigned_and_nowrap. rewrite word.unsigned_and_nowrap. apply Z.land_comm.
```
samuelgruetter, line 30: Instead of writing `rewrite word.unsigned_and_nowrap.` twice, you can also just write `rewrite 2word.unsigned_and_nowrap.`

compiler/src/compiler/FlatToRiscvFunctions.v
```diff
@@ -1848,23 +1851,44 @@ Section Proofs.
...
+          | H: ?x = Syntax.bopname.lts, y: operand |- _ =>
+              destr y; simpl in *; run1det; run1done; fwd; eauto 8 with map_hints
+          | H: ?x = Syntax.bopname.sru, y: operand |- _ =>
+              destr y; simpl in *; run1det; run1done; [rewrite word.sru_ignores_hibits; eauto 8 with map_hints | fwd; eauto 8 with map_hints ]
+          | H: ?x = Syntax.bopname.slu, y: operand |- _ =>
+              destr y; simpl in *; run1det; run1done; [rewrite word.slu_ignores_hibits; eauto 8 with map_hints | fwd; eauto 8 with map_hints ]
+          | H: ?x = Syntax.bopname.srs, y: operand |- _ =>
+              destr y; simpl in *; run1det; run1done; [rewrite word.srs_ignores_hibits; eauto 8 with map_hints | fwd; eauto 8 with map_hints ]
+          | H: ?x = Syntax.bopname.sub, y: operand |- _ =>
+              destr y; simpl in *; [ run1det; run1done | admit ]
```
0adb, line 1878: Currently, trying to use 
```
eapply runsTo_det_step_with_valid_machine; [ assumption
   | simulate';
  	match goal with
  	| |- ?mid = ?RHS => is_evar mid; simpl; reflexivity
  	| |- _ => idtac
  	end
   | intros ].
```
on one of the goals admitted (which are all goals that use InvalidInstruction (-1)) gives a subgoal `False`. One of the hypotheses for this subgoal looks like 
```
(xframe *
         	(ptsto_instr iset (program_base + !pos)
            	(InvalidInstruction (-1)) * emp True) *
         	functions program_base e_pos e_impl * dframe *
         	eq m *
         	word_array
           	(p_sp -
            	!(bytes_per_word * #(Datatypes.length stack_trash)))
           	stack_trash * word_array p_sp frame_trash)%sep
          	initialL_mem
```
so it might not be because of the modulo possibility discussed before (where maybe `InvalidInstruction (-1)` gets transformed into `InvalidInstruction (2^32 - 1)` or similar).

samuelgruetter: Keep in mind that getting a subgoal of `False` does not mean you can't prove it: If you have a contradictory hypothesis (and I think the one you cite above is), you're still good!
Using the lemma `invert_ptsto_instr` (and `ecancel_assumption` do discharge its separation logic assumption), you can get `(Encode.verify (InvalidInstruction (-1)) iset \/ valid_InvalidInstruction (InvalidInstruction (-1)))`. The left-hand side of this `\/` simplifies to `False`, and the right-hand side should imply `0 <= -1 < 2 ^ 32`, so I think you're almost there!

compiler/src/compiler/FlatToRiscvFunctions.v
```diff
@@ -1889,7 +1889,50 @@ Section Proofs.
...
+               stack_trash * word_array p_sp frame_trash)%sep ) with ((
+             (ptsto_instr iset (program_base + !pos)
+                (InvalidInstruction (-1)) * emp True) *
+             functions program_base e_pos e_impl * dframe *
+             eq m *
+             word_array
+               (p_sp -
+                !(bytes_per_word * #(Datatypes.length stack_trash)))
+               stack_trash * word_array p_sp frame_trash * xframe)%sep ) in H12p6p2.
+          2: { rewrite sep_comm with (q := xframe%sep) (p := ((ptsto_instr iset (program_base + !pos)        (InvalidInstruction (-1)) * emp True) *
```
0adb, line 1918: Currently trying to get the hypothesis above in the right shape for `invert_ptsto_instr` because the lemma takes `ptsto_instr ... * ...` and the hypothesis is roughly `xframe * (ptsto_instr ... * ...) * ...`. Did the above `replace` as a first step and stuck trying to prove first the corresponding goal `xframe * (ptso_instr ... * ...) * ... = (ptso_instr ... * ...) * ... * xframe`. The above `rewrite sep_comm with` produces an error similar to `setoid rewrite failed: Unable to satisfy the following constraints: UNDEFINED EVARS:` while switching `p` and `q` gives an unable to unify error. Not sure if I should be using something else besides `sep_comm` or trying to do a different rewrite instead.

samuelgruetter: Yeah that's where `ecancel_assumption` is useful: Using commutativity and associativity of `*` would be too cumbersome, so we prefer to use separation logic canceling for such goals. Try something like this:

```coq
assert (Encode.verify (InvalidInstruction (-1)) iset \/
          valid_InvalidInstruction (InvalidInstruction (-1))). {
  eapply invert_ptsto_instr.
  ecancel_assumption.
}
```
If that doesn't work, replace `ecancel_assumption` by `use_sep_assumption. ecancel.` and try to understand what these tactics do. Or even more manually, to debug: If you get a goal like

```
...
  ======== seps iff ========
...
```
you can use `cancel_seps_at_indices i%nat j%nat` to cancel the i-th clause above `======== seps iff ========` with the j-th clause below.

compiler/src/compiler/FlatToRiscvFunctions.v
```diff
@@ -1877,62 +1877,139 @@ Section Proofs.
...
+          {
+            unfold Encode.verify in H6. simpl in H6. destruct H6. assumption.
+
+          }
+          {
+            unfold valid_InvalidInstruction in H6. fwd. destruct H6p0. destruct H6. destruct H0. simpl. reflexivity.
+          }
+        }
+        contradiction.
+       }
```
0adb, line 1911: The proof Technically succeeds but unsure how to clean it up or make it less bad. Specifically, not sure how to write a match goal on a hypothesis that contains `InvalidInstruction (-1)` without matching on the whole hypothesis exactly.

samuelgruetter: Congrats on the step of getting it to 'work' :tada: :clap: 
Now, to make it less verbose, as you suspect, there's a feature to match on a hypothesis without spelling out everything in it: grep for `context` in the repo, and look for matches that use `context`, probably something like this will help you:
```coq
match goal with
| H: context[InvalidInstruction (-1)] |- _ => ...
end
```
And another hint: `assert False. { proofOfFalse. } contradiction.` can be shortened into `exfalso. proofOfFalse.`

compiler/src/compiler/FlatToRiscvFunctions.v
```diff
@@ -1332,6 +1332,9 @@ Section Proofs.
     + assumption.
   Qed.
 
+
+  Ltac run_and_done := run1det; run1done.
+
```
samuelgruetter, line 1337: Looks like you're not using this, so I'd delete it. Too many indirections make the code harder to read.

compiler/src/compiler/FlatToRiscvFunctions.v
```diff
@@ -1848,23 +1851,71 @@ Section Proofs.
...
+          | H: ?x = Syntax.bopname.mulhuu, y: operand |- _ =>
+              destr y; simpl in *; [ run1det; run1done; rewrite word.mulhuu_simpl; eauto 8 with map_hints  |  ]
+
+          | H: ?x = Syntax.bopname.divu, y: operand |- _ =>
+              destr y; simpl in *; [ run1det; run1done; rewrite word.divu0_simpl; eauto 8 with map_hints | ]
+          | H: ?x = Syntax.bopname.remu, y: operand |- _ =>
+              destr y; simpl in *; [ run1det; run1done; rewrite word.modu0_simpl; eauto 8 with map_hints | ]
+          | H: ?x = Syntax.bopname.eq, y: operand |- _ =>
+              destr y; simpl in *; [ run1det; run1det; run1done; rewrite reduce_eq_to_sub_and_lt; rewrite map.put_put_same; eauto 8 with map_hints |  ]
+          end
```
samuelgruetter, line 1890: This big match contains quite some duplication. Here are some ingredients that could help deduplicate:
* `solve [ tac ]` makes sure that `tac` solves the current goal, i.e. leaves 0 open subgoals
* `tac; [ ]` makes sure that `tac` leaves open exactly one subgoal, otherwise fails
* `tac; [ | ]` makes sure that `tac` leaves open exactly two subgoals, otherwise fails, etc for bigger number of subgoals
* Sometimes, to get the expected subgoal counts, you need parentheses: eg `destruct foo; destruct bar; [ | ]` is the same as `(destruct foo; destruct bar); [ | ]`, but maybe you wanted `destruct foo; (destruct bar; [ | ])`
* `try tac` does nothing if `tac` fails
* `rewrite ?mylemma` rewrites with `mylemma` as many times as possible
* `rewrite !mylemma` rewrites with `mylemma` as many times as possible, but at least once, otherwise fails
* `try rewrite mylemma` rewrites with `mylemma` zero times or once

Using these, you could probably treat many of the operands at once, maybe using something like this:
```coq
all: try (destr y; simpl in *; run1det; run1done; fwd; [ solve [eauto 8 with map_hints] | ]).
```

compiler/src/compiler/FlatToRiscvFunctions.v
```diff
@@ -1848,23 +1851,71 @@ Section Proofs.
...
+        | H : _ |- Encode.verify (InvalidInstruction (-1)) iset \/
+            valid_InvalidInstruction (InvalidInstruction (-1))
+          => eapply invert_ptsto_instr; ecancel_assumption
+          end.
+      all:
+        match goal with
+        | H: Encode.verify (InvalidInstruction (-1)) iset \/
+               valid_InvalidInstruction (InvalidInstruction (-1)) |- _ => exfalso; destruct H; [ unfold Encode.verify in H; simpl in H; destruct H; assumption | unfold valid_InvalidInstruction in H; fwd]
+                                                                             end.
+      all:
```
samuelgruetter, line 1909: `all: (tactic)` blocks like these below are fine, and having `.` in-between them makes the code more debuggable for the next time something breaks.

compiler/src/compiler/FlatToRiscvFunctions.v
```diff
@@ -1848,23 +1851,71 @@ Section Proofs.
...
+        try match goal with
+        | H : _ |- Encode.verify (InvalidInstruction (-1)) iset \/
+            valid_InvalidInstruction (InvalidInstruction (-1))
+          => eapply invert_ptsto_instr; ecancel_assumption
+          end.
+      all:
+        match goal with
+        | H: Encode.verify (InvalidInstruction (-1)) iset \/
+               valid_InvalidInstruction (InvalidInstruction (-1)) |- _ => exfalso; destruct H; [ unfold Encode.verify in H; simpl in H; destruct H; assumption | unfold valid_InvalidInstruction in H; fwd]
+                                                                             end.
```
samuelgruetter, line 1908: How much of these `all:` blocks could you move further up, before doing any `destruct`?

samuelgruetter (commented):
Cool! I added some hints on how to compact it a bit further (but as I already said, don't go tooo far :wink:)

compiler/src/compiler/FlatToRiscvFunctions.v
0adb: Currently I think the all blocks mostly have to happen after both destructs in the SOp case because the InvalidInstruction (-1) only comes out in specific cases of operand being constant and only for certain ops (e.g. mulhuu).

samuelgruetter: I see, makes sense!

samuelgruetter:
superseded by https://github.com/mit-plv/bedrock2/pull/337

PR #315, opened 2023-01-16
More work on elaborator

psvenk:
I have begun to implement records; so far, I have had success (i.e., Coq accepts my definition) in implementing a `PERecord` constructor for unnamed tuples.

For named tuples, how should we store the names in the `expr`s? We would want something like `(fst ({a: 0}, 1)).a` to interpret to `0` while `(fst ({a: 0}, 1)).b` would not be well-typed; however, the elaborator only has access to the type of the left-hand side of the `PEProj` operator (as it is not a `PERecord`) — this suggests that we should store the fact that `a` is a valid key in the type system somehow. Should this be done by implementing named pairs in addition to unnamed pairs, encoding the names in the type system, and having surface syntax of pairs (elaborated to unnamed pairs) and records (elaborated to nested named pairs)?

### DIFF
--- a/PyLevelLang/src/PyLevelLang/Elaborate.v
+++ b/PyLevelLang/src/PyLevelLang/Elaborate.v
@@ -210,24 +210,34 @@ Section WithMap.
         elaborate_proj elaborate G p s
     end.
 
-  Fixpoint elaborate_command (G : tenv) (pc : pcommand) : result command :=
+  Fixpoint elaborate_command (G : tenv) (pc : pcommand)
+    : result (tenv * command) :=
     match pc with
     | PCSkip =>
-        Success CSkip
+        Success (G, CSkip)
     | PCSeq pc1 pc2 =>
-        c1 <- elaborate_command G pc1;;
-        c2 <- elaborate_command G pc2;;
-        Success (CSeq c1 c2)
+        '(G', c1) <- elaborate_command G pc1;;
+        '(G'', c2) <- elaborate_command G' pc2;;
+        Success (G'', CSeq c1 c2)
     | PCLet x p pc =>
         '(existT _ t e) <- elaborate G p;;
         match map.get G x with
         | Some (_, true) =>
             error:("Variable" x "already defined as mutable")
-        | Some (_, false) | None =>
+        | Some (t', false) =>
             (* If already defined as immutable, shadow that variable *)
             let G' := map.put G x (t, false) in
-            c <- elaborate_command G pc;;
-            Success (CLet x e c)
+            '(G'', c) <- elaborate_command G' pc;;
+            (* Keep any changes made to `G` as a result of `c`, but restore `x`
+             * to what it was before *)
+            let G''' := map.put G'' x (t', false) in
+            Success (G''', CLet x e c)
+        | None =>
+            let G' := map.put G x (t, false) in
+            '(G'', c) <- elaborate_command G' pc;;
+            (* Idem *)
+            let G''' := map.remove G'' x in
+            Success (G''', CLet x e c)
         end
     | PCLetMut x p pc =>
         '(existT _ t e) <- elaborate G p;;
@@ -238,15 +248,17 @@ Section WithMap.
             error:("Variable" x "already defined as immutable")
         | None =>
             let G' := map.put G x (t, true) in
-            c <- elaborate_command G pc;;
-            Success (CLetMut x e c)
+            '(G'', c) <- elaborate_command G' pc;;
+            (* `x` goes out of scope *)
+            let G''' := map.remove G'' x in
+            Success (G''', CLetMut x e c)
         end
     | PCGets x p =>
         '(existT _ t e) <- elaborate G p;;
         match map.get G x with
         | Some (t', true) =>
             e' <- enforce_type t' e;;
-            Success (CGets x e')
+            Success (G, CGets x e')
         | Some (_, false) =>
             error:("Variable" x "is not mutable")
         | None =>
@@ -255,16 +267,18 @@ Section WithMap.
     | PCIf p pc1 pc2 =>
         '(existT _ _ e) <- elaborate G p;;
         e' <- enforce_type TBool e;;
-        c1 <- elaborate_command G pc1;;
-        c2 <- elaborate_command G pc2;;
-        Success (CIf e' c1 c2)
+        (* We don't care about changes to G because variables defined in
+         * branches go out of scope at the end of the if statement *)
+        '(_, c1) <- elaborate_command G pc1;;
+        '(_, c2) <- elaborate_command G pc2;;
+        Success (G, CIf e' c1 c2)
     | PCForeach x p pc =>
         '(existT _ t e) <- elaborate G p;;
         match t as t' return expr t' -> _ with
         | TList t => fun e =>
             let G' := map.put G x (t, false) in
-            c <- elaborate_command G' pc;;
-            Success (CForeach x e c)
+            '(_, c) <- elaborate_command G' pc;;
+            Success (G, CForeach x e c)
         | _ => fun _ => error:(e "has type" t "but expected" TList)
         end e
     end.


andres-erbsen:
Yep, the considerations you describe make sense. The elaboration pass that translates records to tuples should return the mapping from record fields from tuple positions (perhaps implied from an alphabetized list of names); though I think that mapping needn't be a part of the type in the target language. My next attempt on top of this code might be something like `elaborate (G : tenv) (p : pexpr) : result {t & expr t * list FieldName }.` Then the projection case would look up the name of the projected field in the list the elaborator returned for that record, but not necessarily record the field names in the target-langauge type. (It might be nice to preserve the field names in target-language code for debugging purposes, though).

PyLevelLang/src/PyLevelLang/Elaborate.v
```diff
@@ -3,155 +3,148 @@ Require Import coqutil.Map.Interface coqutil.Map.SortedListString.
...
   end.
 
 Section WithMap.
   (* abstract all functions in this section over the implementation of the map,
      and over its spec (map.ok) *)
-  Context {tenv: map.map string (type * bool)} {tenv_ok: map.ok tenv}.
+  Context {tenv : map.map string (type * bool)} {tenv_ok: map.ok tenv}.
+
+  Section ElaborateHelpers.
+    Context {elaborate : tenv -> pexpr -> result {t & expr t}}.
```
samuelgruetter, line 20:
```suggestion
    Context (elaborate : tenv -> pexpr -> result {t & expr t}).
```
because you can't infer this argument at use site.

samuelgruetter (commented):
Looks good so far, and I agree with Andres' suggestions, except that `list FieldName` won't be sufficient if we want to support nested records. So instead of returning `list FieldName`, maybe a "pretype" would be better, which is defined like `type`, but additionally has named records? But then, what if the elaborator returns a type and pretype that don't match? So maybe it should return `result {pt: pretype & expr (elaborate_type pt)}`, where `elaborate_type` just replaces named record types by unnamed tuples? But that looks a bit heavyweight... And moreover, for debugging purposes, we'll want to include the string names somewhere in the elaborated `expr` (but without enforcing their consistency in Coq's typesystem). So we could include this name information either in the terms, or actually also in `type`, and then, returning `result {t & expr t}` in the elaborator would again be sufficient, because if `t` is a pair that is the result of translating a record expression, you could retrieve the names from this additional naming information stored in `type`. These are just some thoughts, probably requires some experimentation into different directions to find out what works best.

PyLevelLang/src/PyLevelLang/Language.v
```diff
@@ -13,6 +13,7 @@ Inductive type : Type :=
   | TBool
   | TString
   | TPair (t1 t2 : type)
+  | TPair' (s : string) (t1 t2 : type) (* For records *)
```
samuelgruetter, line 16: Did you consider using one single pair constructor with an `option string` instead of two pair constructors? This could reduce some duplication

psvenk: How about just doing away with unnamed pairs in `expr` and elaborating `PEPair p1 p2` to `EPair "0" p1 (EPair "1" p2 (EConst CEmpty))` (i.e., the same as `PERecord (("0", p1) :: ("1", p2) :: nil)`)? This would also mean removing `PEFst` and `PESnd` in favor of the `0` and `1` projections. This does mean that a pair such as `PEPair (PEConst (CInt 2)) (PEConst (CBool true))` would evaluate to `(2, (true, ()))` and not `(2, true)`, but I imagine that this should not be a big deal.

PyLevelLang/src/PyLevelLang/Language.v
```diff
@@ -122,10 +115,18 @@ Inductive expr : type -> Type :=
...
 Inductive command : Type :=
   | CSkip
   | CSeq (c1 c2 : command)
-  | CLet (t : type) (x : string) (e : expr t) (c : command)
-  | CLetMut (t : type) (l : string) (e : expr t) (c : command)
+  | CLet {t : type} (x : string) (e : expr t) (c : command)
+  | CLetMut {t : type} (l : string) (e : expr t) (c : command)
   | CIf (e : expr TBool) (c1 c2 : command)
-  | CForeach (t : type) (x : string) (e : expr (TList t)) (c : command).
+  | CForeach {t : type} (x : string) (e : expr (TList t)) (c : command).
```
samuelgruetter, line 132: I would add a `| CSet {t : type} (x : string) (e : expr t)`, which requires that `x` has already been declared before with a `CLetMut`, and also disallow `CLetMut` for variables that are already bound, just to make sure we're explicit about declaring each variable and about its scope.

PyLevelLang/src/PyLevelLang/Language.v
```diff
@@ -106,7 +97,9 @@ Inductive pexpr : Type :=
   | PEBinop (po : pbinop) (p1 p2 : pexpr)
   | PEFlatmap (p1 : pexpr) (x : string) (p2 : pexpr)
   | PEIf (p1 p2 p3 : pexpr)
-  | PELet (x : string) (p1 p2 : pexpr).
+  | PELet (x : string) (p1 p2 : pexpr)
+  | PERecord (xs : list (string * pexpr))
```
samuelgruetter, line 101: Do we still need `PERecord`, or could it be just written using `PEBinop` and `POPair`?

PyLevelLang/src/PyLevelLang/Language.v
```diff
@@ -12,9 +12,9 @@ Inductive type : Type :=
   | TInt
   | TBool
   | TString
-  | TPair (t1 t2 : type)
-  | TList (t : type)
-  | TEmpty. (* "Empty" type: its only value should be the empty tuple () *)
+  | TPair (s : string) (t1 t2 : type) (* s = label of t1 in a record *)
```
samuelgruetter, line 15: This is interesting: Without originally intending to enforce with dependent types that record field names matter in types, we now do enforce that, eg you can't assign a variable of type `{ x: Int, y: Int }` to a variable of type `{ a: Int, b: Int }`, which is good.

samuelgruetter (commented):
Looks great, and I think using the default string names "0" and "1" for unnamed tuples so that we don't really need them any more seems to work!

PyLevelLang/src/PyLevelLang/Language.v
psvenk: I realize that the naming scheme I went with is a bit confusing; while `OPair` is used for named pairs (as a representation of records/tuples), `POPair` is for unnamed pairs (which get elaborated into a special case of records). So, `PERecord` is actually more expressive than `POPair`; I kept `POPair` for now because I thought it would be useful to have syntactic sugar for unnamed pairs.

PyLevelLang/src/PyLevelLang/Language.v
psvenk: Yes; as I was writing this, I realized that I could just enforce field names using the type system and took the opportunity.

PyLevelLang/src/PyLevelLang/Language.v
psvenk: I implemented this as `CGets` in b9773bac3b. I kept variable shadowing when an immutable variable shadows another immutable variable or when `PELet` is used, as that could be useful (though I realize that this creates some inconsistency in the syntax and that some languages, such as Haskell, do just fine without it).

samuelgruetter: Here's a little test that would be good to test whether the interpreter behaves as expected in such shadowing cases:

```
let mut a := 1; (let x := 2 in ((let x := 3 in a := 4); let mut res := x + a))
```
When computing `x + a`, `x` needs to refer to the older value (2), so it must *discard* the update that happened inside `(let x := 3 in a := 4)`, but `a` needs to refer to the updated value (4), so it must *keep* the update that happened inside `(let x := 3 in a := 4)`.

PyLevelLang/src/PyLevelLang/Elaborate.v
```diff
@@ -210,24 +210,34 @@ Section WithMap.
         elaborate_proj elaborate G p s
     end.
 
-  Fixpoint elaborate_command (G : tenv) (pc : pcommand) : result command :=
+  Fixpoint elaborate_command (G : tenv) (pc : pcommand)
+    : result (tenv * command) :=
```
samuelgruetter, line 214: I don't think this threading is what you should do in the elaborator. I would want scope to follow the structure of commands, eg in `(let mut x := e in c1); c2`, `x` should not be bound in `c2`. But often, you'd write `let mut x := e in (c1; c2)` to make sure it's also bound in `c2`.

PyLevelLang/src/PyLevelLang/Language.v
hulsemohit: Thanks for pointing this out! I overlooked this issue in my first implementation of `interp_command`, but I think I've fixed it now (the example you gave seems to work as expected).

The way I've now implemented it, each `let`/`let mut` effectively opens a new scope. Lookups (`EVar`/`ELoc`)and assignments (`CGets`) for an identifier `x` always refer to the innermost open scope that contains `x`. When all names `x` go out of scope, it is removed. 

The implementation is simplified since each scope introduces only one variable. In Coq:
```coq
    | CLet x e c1 | CLetMut x e c1 => let l' := interp_command (set_local l x (interp_expr l e)) c1
                                      in match map.get l x with
                                         | None => map.remove l' x
                                         | Some v => set_local l' x (projT2 v)
                                         end
```
(The code for `interp_command` is now at [https://github.com/hulsemohit/bedrock2/tree/pylevel-interp](url); perhaps I should make a pull request?)

samuelgruetter: That snippet looks smart! Maybe even a bit too smart, as in, having it appear in all proofs will complicate the proofs and we should just ban all shadowing? Another idea I had was to split the mutable and immutable variables into two separate environments, but right now I don't think that would help much. Just wanted to mention this because maybe at some point it actually will help? And yes, a pull request would be good!

andres-erbsen: My intuition is in favor of the semantics @hulsemohit proposed; the `in` clause in the above snippet could be probably refactored as `override l' x (map.get l x)` using `override : map.rep -> key -> option value -> map.rep` (or idk what to name this, maybe `set` or `replace`).

samuelgruetter:
To ease collaboration (hopefully?), I merged into master all commits of this PR except the [last one](https://github.com/mit-plv/bedrock2/pull/315/commits/649497112bdb8993dcd08f899c718d9e1a455e1d) -- if you think this one is really what we need, you'll still have to convince me :wink:

PyLevelLang/src/PyLevelLang/Language.v
samuelgruetter: Good point, and it turns out we already have this definition: `Search (map.rep -> _ -> option _ -> map.rep)` returns `map.update`

PyLevelLang/src/PyLevelLang/Elaborate.v
psvenk: Good point. I saw the fact that `PCGets` doesn't take a continuation and for some reason forgot that scope follows the structure of commands; looking back at the code that I wrote, the threading is totally unnecessary and doesn't even nontrivially change the `PCGets` case!

psvenk:
Seeing as most of this PR has been merged and that the lone remaining commit was made in a bout of wrongheadedness (see <https://github.com/mit-plv/bedrock2/pull/315#discussion_r1086170693>), I am closing this PR.

PR #317, opened 2023-01-23
Interpreting commands

hulsemohit:
This should handle the kinds of commands currently supported by the language.
As mentioned in #315, the current way of handling shadowing will make proofs complicated. Should shadowing be disabled completely? I can't think of a way to further simplify the current handling…

If we do not remove shadowing, there's the question of how we handle this situation:
```
let mut x := 0 in (let x := 3 in x := x);
```
 If we do want to allow this, then the LHS of the last assignment should refer to the mutable `x` and the RHS should refer to the immutable `x`. Currently, this would be impossible in the interpreter, since it does not distinguish between mutable and immutable bindings. (Perhaps the right thing to do here is to disallow this in the elaborator.)

We might also want to consider adding some kind of 'print' functionality; there's not a good way to inspect local variables or the environment during interpretation right now. Although `interp_command` returns a list of locals, all local bindings will be removed after execution, and there's real no way to step through the execution while testing.

This pull request should be compatible with #315 (in case it makes more sense to merge that one first).

### DIFF
--- a/LiveVerif/src/LiveVerif/LiveParsing.v
+++ b/LiveVerif/src/LiveVerif/LiveParsing.v
@@ -117,11 +117,11 @@ Infix "||" := expr.lazy_or
 Notation "c ? e1 : e2" := (expr.ite c e1 e2)
   (in custom live_expr at level 13, right associativity, only parsing).
 
-Notation "load1( a )" := (expr.load access_size.one a)
+Notation "load8( a )" := (expr.load access_size.one a)
   (in custom live_expr at level 1, a custom live_expr at level 100, only parsing).
-Notation "load2( a )" := (expr.load access_size.two a)
+Notation "load16( a )" := (expr.load access_size.two a)
   (in custom live_expr at level 1, a custom live_expr at level 100, only parsing).
-Notation "load4( a )" := (expr.load access_size.four a)
+Notation "load32( a )" := (expr.load access_size.four a)
   (in custom live_expr at level 1, a custom live_expr at level 100, only parsing).
 Notation  "load( a )" := (expr.load access_size.word a)
   (in custom live_expr at level 1, a custom live_expr at level 100, only parsing).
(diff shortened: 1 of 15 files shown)


samuelgruetter (commented):
Looks great!

Regarding shadowing: This should be informed by what works best in proofs. If you're eager, you could already try to implement/prove some source-to-source optimizations such as eg constant folding or dead code elimination. The proofs of such optimizations can then inform the design of the interpreter. We're defining a somewhat different language and semantics than FRAP, but maybe [this chapter](https://github.com/achlipala/frap/blob/172716c8ef4a30011399dfc837c3936b28ff1fcc/CompilerCorrectness.v#L232) could still be helpful to consult.

Regarding 'print' functionality: An immediate fix to see result variables would be to run sample programs in a non-empty initial environment, so that at the end, the variables that already were there at the beginning are still present, and their values can be tested with something like

```coq
Goal elaborate_and_interp_command initial_env cmd = [expected; final; locals]). 
Proof. compute. reflexivity. Abort.
```
(where the toplevel `elaborate_and_interp_command` might need to call `map.tuples` on the final locals because sometimes computing through maps will simplify their type and string comparison functions way too much).
As a next step, it might be good to use this strategy to add some tests of the command interpreter.

But you're right that a 'print' functionality would be nice, and I could imagine using a slightly more general notion of "external call" for this (which we also have in bedrock2). You could add a `CExtCall` which takes the string name of an external function, and a list of arguments to pass to the call, and for now, I'd say doesn't expect any return values from the external call, and then add a trace to the interpreter which records all the external calls, so that after running the command interpreter, you see all the print (or other) external calls that were made.

andres-erbsen:
My intuition says that `let mut x := 0 in (let x := 3 in x := x);` should give a type error, specifically `attempt to overwrite immutable variable x`.

hulsemohit:
Closing to move to new pull request.

PR #322, opened 2023-01-31
Proof of proper scope handling in elaborate

psvenk:
Creates `wf`, a "well-formedness" judgement for expressions which can be used to state (in `elaborate_wf`) that `elaborate` correctly handles scope (i.e., an elaborated expression never has free variables). The soundness of the types assigned by `elaborate` is already enforced via the GADT.

The proof of `elaborate_wf` is still not complete. The parts that actually have to do with scope were not too hard to write; however, dealing with `enforce_type` and records has proven difficult. As of now, the remaining `admit`s are for proving that two type variables are equal and for the `PEProj` case. @samuelgruetter / @andres-erbsen, do you have any pointers for how to tackle these?

### DIFF
--- a/PyLevelLang/src/PyLevelLang/Elaborate.v
+++ b/PyLevelLang/src/PyLevelLang/Elaborate.v
@@ -294,7 +294,7 @@ Section WithMap.
     end.
 
   (* Well-formedness judgement for `expr`s, stating that an `expr` has no free
-   * variables *)
+   * variables and variables are used with correct types *)
   Inductive wf : tenv -> forall {t : type}, expr t -> Prop :=
     | wf_EVar G t (x : string) :
         map.get G x = Some (t, false) -> wf G (EVar t x)
--- a/PyLevelLang/src/PyLevelLang/Elaborate.v
+++ b/PyLevelLang/src/PyLevelLang/Elaborate.v
@@ -298,4 +298,228 @@ Section WithMap.
         | _ => fun _ => error:(e "has type" t "but expected" TList)
         end e
     end.
+
+  (* Well-formedness judgement for `expr`s, stating that an `expr` has no free
+   * variables *)
+  Inductive wf : tenv -> forall {t : type}, expr t -> Prop :=
+    | wf_EVar G t (x : string) :
+        map.get G x = Some (t, false) -> wf G (EVar t x)
+    | wf_ELoc G t (x : string) :
+        map.get G x = Some (t, true) -> wf G (ELoc t x)
+    | wf_EConst G {t} (c : const t) : wf G (EConst c)
+    | wf_EUnop G {t1 t2} (o : unop t1 t2) (e : expr t1) :
+        wf G e -> wf G (EUnop o e)
+    | wf_EBinop G {t1 t2 t3} (o : binop t1 t2 t3)
+        (e1 : expr t1) (e2 : expr t2) :
+        wf G e1 -> wf G e2 -> wf G (EBinop o e1 e2)
+    | wf_EFlatmap G {t} (e1 : expr (TList t)) (x : string)
+        (e2 : expr (TList t)) :
+        wf G e1 -> wf (map.put G x (t, false)) e2 -> wf G (EFlatmap e1 x e2)
+    | wf_EIf G {t} (e1 : expr TBool) (e2 e3 : expr t) :
+        wf G e1 -> wf G e2 -> wf G e3 -> wf G (EIf e1 e2 e3)
+    | wf_ELet G {t1 t2} (x : string) (e1 : expr t1) (e2 : expr t2) :
+        wf G e1 -> wf (map.put G x (t1, false)) e2 -> wf G (ELet x e1 e2).
+
+  Lemma elaborate_wf :
+    forall p G t e, elaborate G p = Success (existT expr t e) -> wf G e.
+  Proof.
+    intro p.
+    induction p; intros G t e H; unfold elaborate in H; fold elaborate in H.
+    - (* PEVar x *)
+      destruct (map.get G x) as [[t' [|]] |] eqn : Hmap.
+      + (* Some (t', true) *)
+        injection H as [= H' H]. rewrite H' in H.
+        injection H as H. rewrite <- H.
+        apply wf_ELoc. now rewrite Hmap, H'.
+      + (* Some (t', false) *)
+        injection H as [= H' H]. rewrite H' in H.
+        injection H as [= <-].
+        apply wf_EVar. now rewrite Hmap, H'.
+      + (* None *)
+        discriminate H.
+    - (* PEConst pc *)
+      destruct pc; unfold elaborate_const in H.
+      all: injection H as [= _ H].
+      all: destruct t; try easy.
+      all: repeat injection H as [= <-].
+      all: apply wf_EConst.
+    - (* PESingleton p *)
+      destruct (elaborate G p) as [[t' e'] |] eqn : H'; try easy.
+      inversion H.
+      apply wf_EBinop.
+      * now apply IHp.
+      * apply wf_EConst.
+    - (* PEUnop po p *)
+      destruct po; unfold elaborate_unop in H.
+      all: destruct (elaborate G p) as [[t1 e1] |] eqn : H1; try easy.
+      all: try (
+        destruct (enforce_type _ e1) as [e1' |] eqn : H1'; try easy;
+        inversion H;
+        apply wf_EUnop;
+        apply IHp; rewrite H1;
+        enough (existT expr _ e1 = existT expr _ e1');
+        now rewrite <- H0 || apply enforce_type_eq;
+        destruct t1 || exact H1'
+      ).
+      + (* POLength *)
+        destruct t1; try easy; inversion H; apply wf_EUnop, IHp, H1.
+    - (* PEBinop po p1 p2 *)
+      destruct po; unfold elaborate_binop in H.
+      all: destruct (elaborate G p1) as [[t1 e1] |] eqn : H1; try easy.
+      all: destruct (elaborate G p2) as [[t2 e2] |] eqn : H2; try easy.
+      all: try (
+        destruct (enforce_type _ e1) as [e1' |] eqn : H1'; try easy;
+        destruct (enforce_type _ e2) as [e2' |] eqn : H2'; try easy;
+        inversion H;
+        apply wf_EBinop; (
+          apply IHp1; rewrite H1;
+          enough (existT expr _ e1 = existT expr _ e1');
+          now rewrite <- H0 || apply enforce_type_eq;
+          destruct t1 || exact H1'
+        ) || (
+          apply IHp2; rewrite H2;
+          enough (existT expr _ e2 = existT expr _ e2');
+          now rewrite <- H0 || apply enforce_type_eq;
+          destruct t2 || exact H2'
+        )
+      ).
+      + (* POConcat *)
+        destruct t1; try easy;
+        destruct t2; try easy.
+        * inversion H.
+          apply wf_EBinop.
+          -- now apply IHp1.
+          -- now apply IHp2.
+        * destruct (enforce_type _ e2) as [e2' |] eqn : H2'; try easy.
+          inversion H.
+          apply wf_EBinop.
+          -- now apply IHp1.
+          -- apply IHp2. rewrite H2.
+             enough (existT expr _ e2 = existT expr _ e2').
+             { now rewrite <- H0. }
+             apply enforce_type_eq.
+             ++ admit.
+             ++ exact H2'.
+      + (* POEq *)
+        destruct (enforce_type _ e2) as [e2' |] eqn : H2'; try easy.
+        unfold construct_eq in H.
+        destruct t1; try easy.
+        all:
+          inversion H;
+          apply wf_EBinop; (
+            now apply IHp1
+          ) || (
+            apply IHp2; rewrite H2;
+            enough (existT expr _ e2 = existT expr _ e2');
+            now rewrite <- H0 || apply enforce_type_eq;
+            destruct t2 || exact H2'
+          ).
+      + (* PORepeat *)
+        destruct t1; try easy.
+        destruct (enforce_type _ e2) as [e2' |] eqn : H2'; try easy.
+        inversion H.
+        apply wf_EBinop.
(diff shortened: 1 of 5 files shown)


andres-erbsen:
Here you go :)

```diff
diff --git a/PyLevelLang/src/PyLevelLang/Elaborate.v b/PyLevelLang/src/PyLevelLang/Elaborate.v
index df3f586b2..556cb2c5e 100644
--- a/PyLevelLang/src/PyLevelLang/Elaborate.v
+++ b/PyLevelLang/src/PyLevelLang/Elaborate.v
@@ -13,18 +13,11 @@ Definition enforce_type (t1 : type) {t2 : type} (e : expr t2) : result (expr t1)
 
 (* `enforce_type` preserves equality
  * (`existT` needs to be used for the statement to typecheck) *)
-Lemma enforce_type_eq : forall t t',
-  t = t' -> forall (e : expr t) (e' : expr t'),
+Lemma enforce_type_eq {t t'} (e : expr t) (e' : expr t') :
   enforce_type t' e = Success e' ->
   existT expr t e = existT expr t' e'.
 Proof.
-  intros t t' H1 e e' H2.
-  unfold enforce_type in H2.
-  destruct type_eq_dec as [H3 |]; try easy.
-  unfold cast in H2.
-  injection H2 as [= <-].
-  unfold eq_rect.
-  now destruct H3.
+  cbv [enforce_type]; case type_eq_dec eqn:E; inversion_clear 1; subst; trivial.
 Qed.
 
 Section WithMap.
@@ -391,15 +384,13 @@ Section WithMap.
           -- now apply IHp1.
           -- now apply IHp2.
         * destruct (enforce_type _ e2) as [e2' |] eqn : H2'; try easy.
-          inversion H.
+          inversion_clear H.
+
           apply wf_EBinop.
           -- now apply IHp1.
           -- apply IHp2. rewrite H2.
-             enough (existT expr _ e2 = existT expr _ e2').
-             { now rewrite <- H0. }
-             apply enforce_type_eq.
-             ++ admit.
-             ++ exact H2'.
+             f_equal.
+             apply enforce_type_eq; auto.
       + (* POEq *)
         destruct (enforce_type _ e2) as [e2' |] eqn : H2'; try easy.
         unfold construct_eq in H.
@@ -425,7 +416,6 @@ Section WithMap.
           { now rewrite <- H0. }
           apply enforce_type_eq.
           -- now destruct t2.
-          -- exact H2'.
       + (* POPair *)
         inversion H.
         apply wf_EBinop.
@@ -442,7 +432,6 @@ Section WithMap.
           enough (existT expr _ e2 = existT expr _ e2').
           { now rewrite <- H0. }
           apply enforce_type_eq.
-          -- admit.
           -- exact H2'.
     - (* PEFlatmap p1 x p2 *)
       destruct (elaborate G p1) as [[t1 e1] |] eqn : H1; try easy.
@@ -457,7 +446,6 @@ Section WithMap.
         enough (existT expr t2 e2 = existT expr (TList t1) e2').
         { now rewrite <- H0. }
         apply enforce_type_eq.
-        * admit.
         * exact H2'.
     - (* PEIf p1 p2 p3 *)
       destruct (elaborate G p1) as [[t1 e1] |] eqn : H1; try easy.
@@ -472,7 +460,6 @@ Section WithMap.
         enough (existT expr t1 e1 = existT expr TBool e1').
         { now rewrite <- H0. }
         apply enforce_type_eq.
-        * now destruct t1.
         * exact H1'.
       + apply IHp2, H2.
       + apply IHp3.
@@ -480,7 +467,6 @@ Section WithMap.
         enough (existT expr t3 e3 = existT expr t2 e3').
         { now rewrite <- H0. }
         apply enforce_type_eq.
-        * admit.
         * exact H3'.
     - (* PELet x p1 p2 *)
       destruct (elaborate G p1) as [[t1 e1] |] eqn : H1; try easy.
~~~

File: PyLevelLang/src/PyLevelLang/Elaborate.v [hunk 1]
```diff
@@ -282,4 +299,236 @@ Section WithMap.
         | _ => fun _ => error:(e "has type" t "but expected" TList)
         end e
     end.
+
+  (* Well-formedness judgement for `expr`s, stating that an `expr` has no free
+   * variables *)
```
samuelgruetter, line 304: I think this comment should also highlight the following important detail: `wf` also enforces that the types that the GADT-style typing assigns to variables at use site corresponds to their type at declaration site.

andres-erbsen:
...and for `project` I'd suggest trying a lemma roughly stating that successfully projecting a field of a well-formed expression gives a well-formed expression, perhaps proven by induction on that expression.

psvenk:
I lifted the `elaborate` helper functions to take `expr`s instead of `pexpr`s, which allows `elaborate_wf` to be split into lemmas just as `elaborate` was split into helper functions. However, `elaborate_record_wf` is still unfinished because I couldn't figure out how to state the premise that every expression in `xs` is well-formed. I tried using `List.map`, but I ended up getting a `list Prop` and don't know how to make that into a usable `Prop`.

samuelgruetter:
> I tried using `List.map`, but I ended up getting a `list Prop` and don't know how to make that into a usable `Prop`.

Are you just looking for `List.Forall`?

samuelgruetter:
I guess you've encountered some annoying cases where you know that two expressions are equal, but you can't say it because their types are `expr t` and `expr t'`, and in your proofs, you never proved `t = t'`. Thinking of eg `enforce_type_wf`, where `t = t'` would follow from the `enforce_type t' e = Success e'` hypothesis, or `existT_wf`, where `t=t'` would follow from the `(existT expr t e = existT expr t' e')` hypothesis.

So, what you probably really wanted was this:
```coq
Lemma enforce_type_eq {t t'} (e : expr t) (e' : expr t') :
  enforce_type t' e = Success e' -> t = t' /\ e = e'.
```
But, as you noticed, `e = e'` doesn't typecheck.
However, using a helper lemma and some Ltac, you can get an `Ltac invert_enforce_type` that achieves the same, but in two steps (first proving `t = t'`, then substituting one of them, then proving `e = e'`).
Using this strategy, I think you can get rid of lemmas proving statements like `X preserves Y`, like `enforce_type_wf`, `existT_wf`.

```diff
diff --git a/PyLevelLang/src/PyLevelLang/Elaborate.v b/PyLevelLang/src/PyLevelLang/Elaborate.v
index 420b8065..8356072a 100644
--- a/PyLevelLang/src/PyLevelLang/Elaborate.v
+++ b/PyLevelLang/src/PyLevelLang/Elaborate.v
@@ -21,6 +21,22 @@ Proof.
   cbv [enforce_type]; case type_eq_dec eqn:E; inversion_clear 1; subst; trivial.
 Qed.
 
+Lemma enforce_type_type_eq {t t'} (e : expr t) (e' : expr t') :
+  enforce_type t' e = Success e' -> t = t'.
+Proof.
+  unfold enforce_type. case type_eq_dec eqn:E; inversion_clear 1; subst; trivial.
+Qed.
+
+Ltac invert_enforce_type :=
+  match goal with
+  | H: @enforce_type ?t' ?t ?e = Success ?e' |- _ =>
+      pose proof (enforce_type_type_eq _ _ H);
+      try subst t'; try subst t;
+      inversion H;
+      try subst e'; try subst e;
+      clear H
+  end.
+
 (* `enforce_type` preserves equality
  * (`existT` needs to be used for the statement to typecheck) *)
 Lemma enforce_type_eq {t t'} (e : expr t) (e' : expr t') :
@@ -125,6 +141,8 @@ Section WithMap.
         end e
     end.
 
+  Ltac invert H := inversion H; clear H; subst.
+
   Lemma elaborate_unop_wf (po : punop) {t : type} (e : expr t) : forall G t' e',
     wf G e -> elaborate_unop po e = Success (existT expr t' e') -> wf G e'.
   Proof.
@@ -132,8 +150,9 @@ Section WithMap.
     destruct po; unfold elaborate_unop in H.
     all: try (
       destruct (enforce_type _ e) as [e' |] eqn : H'; try easy;
-      inversion H;
-      apply wf_EUnop; now apply enforce_type_wf with (G := G) in H'
+      repeat invert_enforce_type;
+      invert H; invert H2;
+      now apply wf_EUnop
     ).
     - (* POLength *)
       destruct t; try easy; inversion H; now apply wf_EUnop.
```

psvenk:
> > I tried using `List.map`, but I ended up getting a `list Prop` and don't know how to make that into a usable `Prop`.
> 
> Are you just looking for `List.Forall`?

Yes, that's exactly what I was looking for. Thanks!

psvenk:
When I was finishing 0835e40cfe, I forgot to check with `Guarded.` and it turns out that Coq won't accept my proof of `elaborate_wf` because it doesn't recognize `p` (from the head of `xs`) as a subterm of `xs`. I tried changing the order of induction, but I still can't get Coq to give me an induction hypothesis that gives me information about the head and not only the tail of the list. Is there some other method that I can use to get Coq to give me an appropriate induction hypothesis (to avoid `Fixpoint`) or to convince Coq that the recursive proof is in fact strictly decreasing?

samuelgruetter:
I wouldn't use both `induction` and `Fixpoint` at the same time: Either use `Lemma` and `destruct` and manually make sure you only use the IH on smaller arguments, or use `induction` and only use the IH provided by `induction`. Also, note that using `Fixpoint` together with automation (`eauto`, maybe even `easy`), is very dangerous, because you never know whether they use the IH on a non-smaller argument even though another way that only uses the IH with a smaller argument would be available. So I'd really try to use only `induction`.

psvenk:
> So I'd really try to use only `induction`.

Yes, that is the goal; however, in removing the use of `Fixpoint`, I ran into issues because I couldn't get Coq to generate an appropriate IH.

samuelgruetter:
> Yes, that is the goal; however, in removing the use of `Fixpoint`, I ran into issues because I couldn't get Coq to generate an appropriate IH.

Indeed, the autogenerated induction principle is not strong enough here. You can see this by doing

```coq
About pexpr_ind.
```

and inspecting its type. For instance, the case for `PEUnop` is good:

```coq
(forall (po : punop) (p : pexpr), P p -> P (PEUnop po p)) ->
```

The `P p` corresponds to an IH for the subterm `p`.
However, the case for `PERecord` is not good:

```coq
(forall xs : list (string * pexpr), P (PERecord xs)) ->
```

As you can see, it does not have any IH for the `pexpr`s buried insied `xs`.
There are two ways to fix this:

1) Rewrite the `Inductive pexpr` so that the autogenerated induction principle becomes strong enough. Typically requires replacing `PERecord` by something like `PERecordCons` and `PERecordNil`.
2) Prove your own induction principle. Start by copy-pasting the output of `About pexpr_ind.`, add the IH you need to the `PERecord` case, and prove it as a `Fixpoint` in tactic mode, and then, to use it, you can write `induction p using my_own_pexpr_ind`.

Both "solutions" require you to write code that you'd prefer not to write :stuck_out_tongue: 
Feel free to pick the one that you think sucks less :wink:
If you need help with 2), it might be worth pinging @DIJamner, I think he's excited about and expert in this approach.

psvenk:
> Both "solutions" require you to write code that you'd prefer not to write :stuck_out_tongue:
> Feel free to pick the one that you think sucks less :wink:

In that case, I think the least bad option would be to simply revert 13889e881d and 0835e40cfe and deal with the fact that there isn't full parallelism between `elaborate_record` and the other `elaborate_$x` helper functions; in any case, this gives a notably cleaner implementation in the record case as, by dealing with `pexpr`s directly, it avoids using things like `List.forall` and
```coq
let f '(s, p) :=
  e <- elaborate G p;;
  Success (s, e)
in
xs' <- List.all_success (List.map f xs);;
elaborate_record xs'
```
to formulate intermediate hypotheses and representations.

samuelgruetter:
> least bad option would be to simply revert [13889e8](https://github.com/mit-plv/bedrock2/commit/13889e881de3a76a4618473131002712291cfff4) and [0835e40](https://github.com/mit-plv/bedrock2/commit/0835e40cfe600af587d9d1edd1fcbd2c11348915) 

Didn't you already have the same guardedness problem before these two commits (maybe without noticing)?

psvenk:
> Didn't you already have the same guardedness problem before these two commits (maybe without noticing)?

No; `elaborate_wf` is defined as a `Fixpoint` and `induction` is also used, but because I am recursing on `p`, which is a part of the head of the input list `xs`, Coq is convinced that the one recursive call is strictly decreasing. (However, after these two commits, this is not the case, as I would be recursing on a part of the head of a new list `xs'` which was derived from `xs`.)

Of course, a `Fixpoint` proof is still probably not the best design, but anyway writing and using a custom inductive hypothesis would be easier in this case because there aren't as many layers of indirection.

samuelgruetter:
Ok, so I just resolved my surprise regarding why this works: Coq's termination checker unfolds as many definitions as needed until it is convinced that all recursive calls are decreasing, and in the case of `elaborate_wf`, it actually unfolds the proof body of the auto-generated induction principle `pexpr_ind` used by `induction` (you can verify this claim by doing `About pexpr_ind.`, copy-pasting its type into the type of a new `Axiom my_pexpr_ind`, and then do `induction p using my_pexpr_ind`, and see how the termination checker fails on the `Qed`). 
Sometimes, relying on too much unfolding in the termination checker can make the `Qed` slow, but here it seems still fast, so as long as it works, I guess this solution is fine, given how bad the other two solutions are.

samuelgruetter:
So, since you're already getting an IH using `Fixpoint`, you could now also just `destruct p` instead of `induction p`, and replace  `now apply IHp` by `eapply elaborate_wf. eassumption.`, which might seem a bit cleaner depending on the reader's taste... :wink:

psvenk:
> So, since you're already getting an IH using `Fixpoint`, you could now also just `destruct p` instead of `induction p`, and replace `now apply IHp` by `eapply elaborate_wf. eassumption.`, which might seem a bit cleaner depending on the reader's taste... :wink:

I just now tried that, but this turns out to be not so simple because the `H2` has type `elaborate_record elaborate G xs = Success (existT expr t2 e2)`, so the left-hand side isn't in terms of `elaborate`. So, I don't think refactoring the proof to avoid use of `induction` is worth the cost.

samuelgruetter:
Cool, so now that `wf_command` and `elaborate_command_wf` are in, this might be a good milestone for merging! Or is there anything related that you still wanted to add?

PR #323, opened 2023-02-01
Add insertion sort and linked list proof drafts

tchomphoochan:
@samuelgruetter Here are the proofs we were working on today, cleaned up a little bit. These should serve well as test cases when developing automation.


samuelgruetter:
My latest changes broke the files of this PR, and I wanted to see the goals that come up, so I already merged it into master, and fixed what I broke.

samuelgruetter:
(it appears as "closed" here on GitHub instead of "merged" because I cherry-picked the commit, so it gets a different hash)

PR #324, opened 2023-02-02
Add notations for commands and types. Also create sample programs

Goel25:
This has the completed notations for commands and types. I've also written a simple `isSquare` program to demonstrate its usability. I'm currently working on proving the correctness of it.

Let me know if any changes should be made.

### DIFF
--- a/PyLevelLang/src/PyLevelLang/Interpret.v
+++ b/PyLevelLang/src/PyLevelLang/Interpret.v
@@ -125,7 +125,11 @@ Section WithMap.
                                          end
     | CGets x e => set_local l x (interp_expr l e)
     | CIf e c1 c2 => if interp_expr l e then interp_command l c1 else interp_command l c2
-    | CForeach x e c1 =>
-        map.remove (fold_left (fun l' v => interp_command (set_local l' x v) c1) (interp_expr l e) l) x
+    | CForeach x e c1 => let l' := fold_left (fun l' v => interp_command (set_local l' x v) c1) (interp_expr l e) l
+                         in match map.get l x with
+                            | None => map.remove l' x
+                            | Some v => set_local l' x (projT2 v)
+                            end
+
     end.
 End WithMap.
--- a/PyLevelLang/src/PyLevelLang/Interpret.v
+++ b/PyLevelLang/src/PyLevelLang/Interpret.v
@@ -118,18 +118,11 @@ Section WithMap.
     match c with
     | CSkip => l
     | CSeq c1 c2 => interp_command (interp_command l c1) c2
-    | CLet x e c1 | CLetMut x e c1 => let l' := interp_command (set_local l x (interp_expr l e)) c1
-                                      in match map.get l x with
-                                         | None => map.remove l' x
-                                         | Some v => set_local l' x (projT2 v)
-                                         end
+    | CLet x e c1 | CLetMut x e c1 => let l' := interp_command (set_local l x (interp_expr l e)) c1 in
+                                      map.update l' x (map.get l x)
     | CGets x e => set_local l x (interp_expr l e)
     | CIf e c1 c2 => if interp_expr l e then interp_command l c1 else interp_command l c2
-    | CForeach x e c1 => let l' := fold_left (fun l' v => interp_command (set_local l' x v) c1) (interp_expr l e) l
-                         in match map.get l x with
-                            | None => map.remove l' x
-                            | Some v => set_local l' x (projT2 v)
-                            end
-
+    | CForeach x e c1 => let l' := fold_left (fun l' v => interp_command (set_local l' x v) c1) (interp_expr l e) l in
+                         map.update l' x (map.get l x)
     end.
 End WithMap.
--- a/PyLevelLang/src/PyLevelLang/Interpret.v
+++ b/PyLevelLang/src/PyLevelLang/Interpret.v
@@ -125,6 +125,7 @@ Section WithMap.
                                          end
     | CGets x e => set_local l x (interp_expr l e)
     | CIf e c1 c2 => if interp_expr l e then interp_command l c1 else interp_command l c2
-    | CForeach x e c1 => fold_left (fun l' v => interp_command (set_local l' x v) c1) (interp_expr l e) l
+    | CForeach x e c1 =>
+        map.remove (fold_left (fun l' v => interp_command (set_local l' x v) c1) (interp_expr l e) l) x
     end.
 End WithMap.
--- a/PyLevelLang/src/PyLevelLang/Notations.v
+++ b/PyLevelLang/src/PyLevelLang/Notations.v
@@ -55,12 +55,12 @@ Notation "'for' x 'in' p : c 'end'"  := (PCForeach x p c)
    (in custom py_comm at level 80, p custom py_expr, c custom py_comm) : pylevel_scope.
 
 (* Type parsing (Types are prefixed with @ so they do not become keywords and pollute the namespace *)
-Notation "@ 'int'"         := TInt (in custom py_type at level 99) : pylevel_scope.
-Notation "@ 'bool'"        := TBool (in custom py_type at level 99) : pylevel_scope.
-Notation "@ 'string'"      := TString (in custom py_type at level 99) : pylevel_scope.
-Notation "@ 'pair' s x y"  := (TPair s%string x y) (in custom py_type at level 99) : pylevel_scope.
-Notation "@ 'empty'"       := TEmpty (in custom py_type at level 99) : pylevel_scope.
-Notation "@ 'list' t"      := (TList t) (in custom py_type at level 99) : pylevel_scope.
+Notation Int    := TInt.
+Notation Bool   := TBool.
+Notation String := TString.
+Notation Pair   := TPair.
+Notation Unit   := TEmpty.
+Notation List   := TList.
 
 
 (* Expression parsing *)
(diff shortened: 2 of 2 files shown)


PyLevelLang/src/PyLevelLang/Interpret.v
```diff
@@ -125,6 +125,7 @@ Section WithMap.
                                          end
     | CGets x e => set_local l x (interp_expr l e)
     | CIf e c1 c2 => if interp_expr l e then interp_command l c1 else interp_command l c2
-    | CForeach x e c1 => fold_left (fun l' v => interp_command (set_local l' x v) c1) (interp_expr l e) l
+    | CForeach x e c1 =>
+        map.remove (fold_left (fun l' v => interp_command (set_local l' x v) c1) (interp_expr l e) l) x
```
samuelgruetter, line 129: Instead of always doing `map.remove`, shouldn't this use the same pattern as `CLet`/`CLetMut` above? (but preferably using `map.update` for this)

PyLevelLang/src/PyLevelLang/Notations.v
```diff
@@ -24,124 +24,206 @@ Coercion PCInt : Z >-> pconst.
...
+Notation "'for' 'each' x 'in' p : c 'end'"  := (PCForeach x p c)
+   (in custom py_comm at level 80, p custom py_expr, c custom py_comm) : pylevel_scope.
+
+(* Type parsing (Types are prefixed with @ so they do not become keywords and pollute the namespace *)
+Notation "@ 'int'"         := TInt (in custom py_type at level 99) : pylevel_scope.
+Notation "@ 'bool'"        := TBool (in custom py_type at level 99) : pylevel_scope.
+Notation "@ 'string'"      := TString (in custom py_type at level 99) : pylevel_scope.
+Notation "@ 'pair' s x y"  := (TPair s%string x y) (in custom py_type at level 99) : pylevel_scope.
+Notation "@ 'empty'"       := TEmpty (in custom py_type at level 99) : pylevel_scope.
+Notation "@ 'list' t"      := (TList t) (in custom py_type at level 99) : pylevel_scope.
```
samuelgruetter, line 70:
```suggestion
Notation Int    := TInt.
Notation Bool   := TBool.
Notation String := TString.
Notation Pair   := TPair.
Notation Unit   := TEmpty.
Notation List   := TList.
```

These `@` don't look so appealing to me... I'd prefer using so-called *abbreviations*, ie notations whose left-hand side is not between quotes, but just an identifier. Their advantage is that they don't generate keywords, and they only pollute the namespace if you `Import` this file, so if you use uppercase letters for the types, there should not be too much potential for confusion (except that `String` is the name of the cons constructor of `string`, but that one is rarely used).
The disadvantage of abbreviations is that they don't support scopes nor custom entries, but I think using Gallina (constr) syntax for types would be fine.

PyLevelLang/src/PyLevelLang/Notations.v
```diff
@@ -24,124 +24,206 @@ Coercion PCInt : Z >-> pconst.
...
+Notation "x"             := x (in custom py_comm at level 0, x constr at level 0) : pylevel_scope.
+Notation "x"             := x (in custom py_type at level 0, x constr at level 0) : pylevel_scope.
 
 (* TODO Allow string constants (which are different from vars) *)
 
+(* Command parsing *)
+Notation "'skip'"                     := (PCSkip)
+   (in custom py_comm at level 0) : pylevel_scope.
+Notation "c ; "                    := c
+   (in custom py_comm at level 90) : pylevel_scope.
```
samuelgruetter, line 50: Why is this notation needed? I would say that the last command in a series of commands does not need to be followed by a semicolon

PyLevelLang/src/PyLevelLang/Notations.v
```diff
@@ -24,124 +24,206 @@ Coercion PCInt : Z >-> pconst.
...
+   reflexivity. Abort.
+   Goal <{ "_" <- [2,4] :: nil(@list @int) }> = PCGets "_"
+      (PEBinop POCons (PEBinop POCons 2 (PESingleton 4)) (PCNil (TList TInt))).
+   reflexivity. Abort.
 
+   (* TODO Is this how the precedence should be for let? *)
+   Goal <{ let "x" := 3 + 4 in "y" <- "x"+1; "z" <- 5+"x" }> =
+      (PCLet "x" (PEBinop POPlus 3 4)
+         (PCSeq (PCGets "y" (PEBinop POPlus "x" 1))
+                (PCGets "z" (PEBinop POPlus 5 "x")))).
```
samuelgruetter, line 194: This example looks good, but I'd also add an example with two nested `let`s

PyLevelLang/src/PyLevelLang/Notations.v
```diff
@@ -24,124 +24,206 @@ Coercion PCInt : Z >-> pconst.
...
-Notation "'for' 'each' x 'in' p : c"  := (PCForeach x p c) (in custom pylevel at level 99) : pylevel_scope.
-*)
+   Goal <{ (let mut "x" := 3 in "y" <- "x" + 1);
+           "x" <- "y" * 2;
+           skip }> =
+           PCSeq (
+             PCSeq (PCLetMut "x" 3 (PCGets "y" (PEBinop POPlus "x" 1)))
+                   (PCGets "x" (PEBinop POTimes "y" 2))
+                 )
+             PCSkip.
```
samuelgruetter, line 204: `c1; c2; c3` should be `PCSeq c1 (PCSeq c2 c3)`, ie it should be right-associative.

PyLevelLang/src/PyLevelLang/Notations.v
```diff
@@ -24,124 +24,206 @@ Coercion PCInt : Z >-> pconst.
...
+   (in custom py_comm at level 90, c1 custom py_comm, c2 custom py_comm) : pylevel_scope.
+Notation "'let' x := p 'in' c"        := (PCLet x p c)
+   (in custom py_comm at level 100, p custom py_expr, c custom py_comm) : pylevel_scope.
+Notation "'let' 'mut' x := p 'in' c"    := (PCLetMut x%string p c)
+   (in custom py_comm at level 100, p custom py_expr, c custom py_comm) : pylevel_scope.
+Notation "x <- p"                     := (PCGets x p)
+   (in custom py_comm at level 50, p custom py_expr) : pylevel_scope.
+Notation "'if' p 'then' c1 'else' c2 'end'" := (PCIf p c1 c2)
+   (in custom py_comm at level 80, p custom py_expr, c1 custom py_comm, c2 custom py_comm) : pylevel_scope.
+Notation "'for' 'each' x 'in' p : c 'end'"  := (PCForeach x p c)
```
samuelgruetter, line 61:
```suggestion
Notation "'for' x 'in' p : c 'end'"  := (PCForeach x p c)
```
I think the `each` is superfluous

samuelgruetter (commented):
Looks great! A few minor suggestions below:

PyLevelLang/src/PyLevelLang/Notations.v
```diff
@@ -24,124 +24,204 @@ Coercion PCInt : Z >-> pconst.
...
-   Goal <{ 1 :: 2 :: [3, 4] }> = PEBinop POCons 1 (PEBinop POCons 2 (PEBinop POCons 3 (PESingleton 4))).
+   Goal <{ "_" <- true }> = PCGets "_" (PEConst (PCBool true)).
+   reflexivity. Abort.
+
+   Goal <{ "_" <- 1 :: 2 :: [3, 4] }> = PCGets "_"
+      (PEBinop POCons 1 (PEBinop POCons 2 (PEBinop POCons 3 (PESingleton 4)))).
    reflexivity. Abort.
-End Tests.
 
+   Goal <{ "_" <- 3 :: 4 :: nil(Int) }> = PCGets "_"
```
samuelgruetter, line 169: Not sure whether `nil(Int)` is good, it looks a bit too much like function application... Or did we actually discuss this in a meeting? What were the considerations? And what about `nil[Int]` or `nil<Int>`?

PyLevelLang/src/PyLevelLang/Notations.v
```diff
@@ -24,124 +24,204 @@ Coercion PCInt : Z >-> pconst.
...
+   reflexivity. Abort.
+   Goal <{ "_" <- true :: false :: nil(Bool) }> = PCGets "_"
+      (PEBinop POCons true (PEBinop POCons false (PCNil TBool))).
+   reflexivity. Abort.
 
+   (* TODO The notation for nil binds too tightly...
+   Goal <{ "_" <- [2,4] :: nil(List Int) }> = PCGets "_"
+      (PEBinop POCons (PEBinop POCons 2 (PESingleton 4)) (PCNil (TList TInt))).
+   reflexivity. Abort.
+    *)
```
samuelgruetter, line 180: I think you could get this example to work by completely eliminating `Custom Entry py_type`, and just say that the argument of `nil` is a `constr`.

PyLevelLang/src/PyLevelLang/Interpret.v
```diff
@@ -125,6 +125,11 @@ Section WithMap.
                                          end
     | CGets x e => set_local l x (interp_expr l e)
     | CIf e c1 c2 => if interp_expr l e then interp_command l c1 else interp_command l c2
-    | CForeach x e c1 => fold_left (fun l' v => interp_command (set_local l' x v) c1) (interp_expr l e) l
+    | CForeach x e c1 => let l' := fold_left (fun l' v => interp_command (set_local l' x v) c1) (interp_expr l e) l
+                         in match map.get l x with
+                            | None => map.remove l' x
+                            | Some v => set_local l' x (projT2 v)
+                            end
```
samuelgruetter, line 132: can you replace this (and above too) by `map.update l' x (map.get l x)`?

PR #325, opened 2023-02-11
Constant folding and branch elimination

hulsemohit:
Reimplementation of optimizations with `atom` instead of the altered `const`.
After rewriting using the `as_const` idea we came up with, the previous issues with the proof of branch elimination and were resolved, so the proofs are complete too.

### DIFF
--- a/PyLevelLang/src/PyLevelLang/Optimize.v
+++ b/PyLevelLang/src/PyLevelLang/Optimize.v
@@ -1,0 +1,170 @@
+Require Import PyLevelLang.Language.
+Require Import PyLevelLang.Interpret.
+Require Import coqutil.Map.Interface coqutil.Map.SortedListString.
+Require Import Coq.Lists.List.
+From Coq Require Import FunctionalExtensionality.
+
+Local Open Scope Z_scope.
+
+Section WithMap.
+  Context {locals: map.map string {t & interp_type t}} {locals_ok: map.ok locals}.
+
+  Fixpoint listify {t : type} (l : list (expr t)) : expr (TList t) :=
+    match l with
+    | nil => EAtom (ANil t)
+    | x :: xs => EBinop (OCons t) x (listify xs)
+    end.
+
+  Lemma listify_correct (t : type) (l : locals) (xs : list (expr t)) :
+    interp_expr l (listify xs) = map (interp_expr l) xs.
+  Proof.
+    induction xs; try easy.
+    simpl.
+    rewrite IHxs.
+    reflexivity.
+  Qed.
+
+  Fixpoint reify (t : type) : interp_type t -> expr t :=
+    match t in type return interp_type t -> expr t with
+    | TInt => fun n => EAtom (AInt n)
+    | TBool => fun b => EAtom (ABool b)
+    | TString => fun s => EAtom (AString s)
+    | TPair s t1 t2 => fun p => EBinop (OPair s t1 t2) (reify t1 (fst p)) (reify t2 (snd p))
+    | TEmpty => fun _ => EAtom AEmpty
+    | TList t => fun l => listify (map (reify t) l)
+    end.
+
+  Lemma reify_correct (t : type) (l : locals) (c : interp_type t) : interp_expr l (reify t c) = c.
+  Proof.
+    induction t; intros; try easy.
+    - destruct c. simpl. rewrite IHt1. rewrite IHt2. reflexivity.
+    - destruct c. reflexivity.
+    - induction c.
+      + reflexivity.
+      + simpl. rewrite IHt. simpl in IHc. rewrite IHc. reflexivity.
+  Qed.
+
+  Fixpoint as_const {t : type} (e : expr t) : option (interp_type t) :=
+    match e with
+    | EAtom a => Some (interp_atom a)
+    | EUnop o e1 => match as_const e1 with
+                    | Some c1 => Some (interp_unop o c1)
+                    | _ => None
+                    end
+    | EBinop o e1 e2 => match as_const e1, as_const e2 with
+                        | Some c1, Some c2 => Some (interp_binop o c1 c2)
+                        | _, _ => None
+                        end
+    | _ => None
+    end.
+
+  Lemma as_const_correct {t : type} (l : locals) (e : expr t) :
+    forall c, as_const e = Some c -> interp_expr l e = c.
+  Proof.
+    generalize dependent l.
+    induction e; intros; try easy; simpl in H.
+    - injection H as H. apply H.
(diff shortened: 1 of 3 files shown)


PyLevelLang/src/PyLevelLang/Optimize.v
```diff
@@ -0,0 +1,170 @@
...
+
+  Fixpoint constant_folding {t : type} (e : expr t) : expr t :=
+    match e with
+    | EVar _ x => EVar _ x
+    | ELoc _ x => ELoc _ x
+    | EAtom a => EAtom a
+    | EUnop o e1 => 
+        let e1' := constant_folding e1 in 
+        match as_const e1' with
+        | Some c => reify _ (interp_unop o c)
```
samuelgruetter, line 82: I think you already pointed this out yourself before implementing it: There's a lot of converting back and forth here (basically three times). And there's also some duplication between `as_const` and `constant_folding`, and if we have big constants, this might even lead to some performance issues.
So, how about trying out this alternative type signature:
```coq
Fixpoint constant_folding {t : type} (e : expr t) : expr t * option (interp_type t) :=
```
The idea would be that you only traverse the term once, ie don't use `as_const` or `reify` here any more (but keep them around because they are generally useful functions), and `constant_folding` always returns the optimized `expr t`, and if it is a constant, it also returns a `Some`, else a `None`.

PyLevelLang/src/PyLevelLang/Optimize.v
```diff
@@ -0,0 +1,291 @@
...
+        let (e1', v) := constant_folding' e1 in
+        match v with
+        | Some v' => let r := interp_unop o v' in (reify _ r, Some r)
+        | _ => (EUnop o e1', None)
+        end
+    | EBinop o e1 e2 =>
+        let (e1', v1) := constant_folding' e1 in
+        let (e2', v2) := constant_folding' e2 in
+        match v1, v2 with
+        | Some v1', Some v2' => let r := interp_binop o v1' v2' in (reify _ r, Some r)
```
samuelgruetter, line 88: Still a bit bothered that you need to call `reify` here. Did yesterday's discussion with Dustin lead to a solution? Or do you think using a non-recursive function with the following signature could help?
```coq
Definition apply_binop_to_reified_consts{t1 t2 t3}
  (op: binop t1 t2 t3)(c1: expr t1)(c2: expr t2): expr t3
```

hulsemohit: We considered essentially the same idea (handling each binop separately). We might also need somewhat smarter ways to check if something is a constant (for example, to fold something like the length of a range) and refrain from unfolding sometimes (for ranges). I'll try to work on implementing and proving this.

PyLevelLang/src/PyLevelLang/Optimize.v
```diff
@@ -70,6 +71,112 @@ Section WithMap.
...
+        | ELoc (TPair s t1 t2) x => EUnop (OFst _ _ _) (ELoc _ x)
+        | @EAtom (TPair s t1 t2) a => EUnop (OFst _ _ _) (EAtom a)
+        | @EUnop _ (TPair s t1 t2) o' e1 => EUnop (OFst _ _ _) (EUnop o' e1)
+        | @EIf (TPair s t1 t2) e1 e2 e3 => EUnop (OFst _ _ _) (EIf e1 e2 e3)
+        | @ELet _ (TPair s t1 t2) x e1 e2 => EUnop (OFst _ _ _) (ELet x e1 e2)
+        | @EBinop _ _ (TPair s t1 t2) o1 e1 e2 => cheat
+        | _ => tt
+         end
+    | OSnd _ _ _ => cheat
+    end.
```
hulsemohit, line 154: I have been trying to factor out unop folding, and as we discussed in the last meeting, it has been difficult to get all the GADT stuff to type-check.  One approach we considered was adding a return clause to the matches so that "impossible" branches return a unit. There still seem to be a few issues with this though (I still can't get it to type check in the Binop case (lines 133 and 150 of this commit)), and it requires a lot of cases (and will be much worse when it comes to binop folding).
We were wondering whether @andres-erbsen had any suggestions?

hulsemohit: We also considered writing the return type of the whole function in a separate definition and handle all cases there. After experimenting a little, I'm not confident that this will help either...

andres-erbsen:
See https://github.com/andres-erbsen/bedrock2/commit/80ca70b428edd149571b722f7dfb4e7fe080ec3f for how to hack through it with the current definitions. Keeping going this way still creates more cases than necessary. See https://github.com/andres-erbsen/bedrock2/commit/afaf70bf93b040eeb0e5505b430f6d48cf33a819 for how I would factor the ~same functionality. (But if you plan to also propagate constants through `ELet` and such, you probably won't want to keep using `fold_expr`).

hulsemohit:
I have now fixed the type of flatmap to allow changing the type of the list—surprisingly, no changes were needed to accommodate this in the elaborator or interpreter. As expected, this allowed defining flatmap-of-flatmap optimizations without the use of proof-mode.

hulsemohit:
There was also a bug in the interpreter where it would evaluate the binary operation OLess as "<=" instead of "<" (I've now fixed this).

PyLevelLang/src/PyLevelLang/Optimize.v
```diff
@@ -62,34 +68,31 @@ Section WithMap.
...
 
   Definition constfold_head {t} (e : expr t) : expr t :=
-    (* This is constant folding. If you don't want to do it for lists,
-    * strings, and pairs, match on a bool here. *)
-    match as_const e with
-    | Some v => reify _ v
+    match t with
+    | TInt | TBool | TString | TEmpty => match as_const e with
+                                         | Some v => reify _ v
+                                         | _ => e
```
andres-erbsen, line 83: I wonder whether we need this conditional in addition to using `invert_atom` in `as_const`.

PyLevelLang/src/PyLevelLang/Optimize.v
```diff
@@ -114,12 +117,15 @@ Section WithMap.
     | o => EUnop o
     end e).
 
+  (*
```
andres-erbsen, line 123: It'd be nice to keep this test working. Would it be enough to have `eLength` return `option Z` and then reify it?

andres-erbsen:
Are the kami/coqutil changes related? if not, they should not be committed. (Also there are merges of master into this branch, so I suggest we squash on merge).

andres-erbsen:
Apparently one needs to click "ready for review" before "merge".

@DIJamner I did it

PR #329, opened 2023-02-22
adapt to coq/coq#17281

andres-erbsen:

PR #331, opened 2023-02-23
Revert "experiment: run limited make on PRs that only change PyLevelLang

andres-erbsen:
This reverts commit 7c572726f2689e970627a290c51147d520208ee0. See https://github.com/mit-plv/bedrock2/pull/329


samuelgruetter:
I agree that this is the way to fix the build failures we've seen on your previous pull requests. Sorry for the trouble that this commit caused.
It seems that this `changed-files` action is still trying to figure out how to get submodules right, there was a change [just last week](https://github.com/tj-actions/changed-files/pull/980), and I'm not too keen on using such bleeeeding edge software :wink:

PR #335, opened 2023-03-02
Add proof for isSquare program correctness

Goel25:
I have completed the proof that the sample `isSquare` program is correct. The proof is a bit tedious and has some long lemmas, so let me know if there are any improvements I should make.

### DIFF
--- a/PyLevelLang/src/PyLevelLang/Notations.v
+++ b/PyLevelLang/src/PyLevelLang/Notations.v
@@ -12,6 +12,7 @@ Declare Scope pylevel_scope.
 Declare Custom Entry py_expr.
 Declare Custom Entry py_comm.
 
+
 Local Open Scope Z_scope.
 Local Open Scope string_scope.
 Local Open Scope list_scope.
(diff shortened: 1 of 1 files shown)


PyLevelLang/src/PyLevelLang/Notations.v
```diff
@@ -226,93 +227,505 @@ End Tests.
...
+     for "x" in range(0, "n"+1):
+         if "x"*"x" == "n"
+           then "ans" <- true
+           else skip
+           end
+     end
+     }>.
+
+  Definition isSquare_elaborated (n : Z) : command :=
+     Eval simpl in (match (elaborate_command (map.of_list [("ans", (TBool, true))]) (isSquare n)) with
```
andres-erbsen, line 247:
```suggestion
     Eval cbv in (match (elaborate_command map.empty (isSquare n)) with
```

PyLevelLang/src/PyLevelLang/Notations.v
```diff
@@ -226,93 +227,505 @@ End Tests.
...
+
+
+Section Square_Section.
   Instance tenv : map.map string (type * bool) := SortedListString.map _.
   Instance tenv_ok : map.ok tenv := SortedListString.ok _.
 
-  Instance locals : map.map string {t & interp_type t} := SortedListString.map _.
-  Instance locals_ok : map.ok locals := SortedListString.ok _.
+  Definition isSquare (n : Z) : pcommand := <{
+     let "n" := n in
```
andres-erbsen, line 237:
```suggestion
     let "n" := n in
     let "ans" := true in
```

PyLevelLang/src/PyLevelLang/Notations.v
```diff
@@ -226,93 +227,505 @@ End Tests.
...
+         if "x"*"x" == "n"
+           then "ans" <- true
+           else skip
+           end
+     end
+     }>.
+
+  Definition isSquare_elaborated (n : Z) : command :=
+     Eval simpl in (match (elaborate_command (map.of_list [("ans", (TBool, true))]) (isSquare n)) with
+                    | Success x => x
```
andres-erbsen, line 248: @samuelgruetter do we have a `of_Success` like `of_Some` (with a dependent return type)?

PyLevelLang/src/PyLevelLang/Notations.v
```diff
@@ -226,93 +227,505 @@ End Tests.
...
+           rewrite map.remove_put_diff by congruence.
+           rewrite map.remove_empty.
+           reflexivity.
+   Qed.
+
+End WithMap.
+
+
+
+Section Examples.
```
andres-erbsen, line 650: Do we really need to have examples that do not start in the empty environment?

PyLevelLang/src/PyLevelLang/Notations.v
```diff
@@ -226,93 +227,505 @@ End Tests.
...
+
+  Definition isSquare_elaborated (n : Z) : command :=
+     Eval simpl in (match (elaborate_command (map.of_list [("ans", (TBool, true))]) (isSquare n)) with
+                    | Success x => x
+                    | _ => _
+                    end
+        ).
+End Square_Section.
+
+
```
andres-erbsen, line 254: The proofs below are a useful sanity-check. I am not sure whether we should keep updating them long-term. @samuelgruetter WDYT is a good plan here?

PyLevelLang/src/PyLevelLang/Notations.v
```diff
@@ -226,93 +227,505 @@ End Tests.
 
 Unset Printing All.
```
DIJamner, line 228: I think we should remove this from the committed code.

DIJamner:
I can understand why `isSquare` and the `Examples` section started out in this file, but we should move them into a separate file from the notation definitions and their dedicated tests.

PyLevelLang/src/PyLevelLang/Notations.v
DIJamner: I couldn't find this function, so I made a coqutil PR for it: mit-plv/coqutil/pull/86. @samuelgruetter: if this function doesn't exist in another place, please either merge it or let me know and I will.

PyLevelLang/src/PyLevelLang/Notations.v
DIJamner: Responding since I'm moving to being more involved with this part: I like having these examples committed (albeit preferably in a separate file), and I think the maintenance burden will be low.

DIJamner:
I agree that the situation with the long lemmas should be able to be improved. However, I think once we've addressed Andres' comments and separated `isSquare` and the rest into another file, we should merge this, and we can leave improving the proofs to a future PR. I'll take a more detailed look at the proofs early next week to give you some feedback and suggestions.

Goel25:
I have moved the examples to a separate file (for now I've just called it `SamplePrograms.v`) and fixed some of the suggestions you've given. Is there anything else that should be changed?

PR #340, opened 2023-03-16
[in progress] working on integrating UseImmediate optimization into Pipeline.v

0adb:
FlatImp currently has an SOp variant in its statement datatype that could be used to compile to RISC-V immediate instructions in certain cases. Optimization to do this has been written in UseImmediate.v but proof of correctness not yet integrated into main pipeline. Example has been made in end2end/src/end2end/End2EndLightbulb.v in module PrintProgram' . Difference between RISC-V assembly produced without UseImmediate and with UseImmediate at https://www.diffchecker.com/KarK1dqd/.

### DIFF
--- a/compiler/src/compiler/Pipeline.v
+++ b/compiler/src/compiler/Pipeline.v
@@ -51,8 +51,8 @@ Require Import coqutil.Tactics.autoforward.
 Require Import compiler.FitsStack.
 Require Import compiler.LowerPipeline.
 Require Import bedrock2.WeakestPreconditionProperties.
-Require Import compiler.UseImmediate.
 Require Import compiler.UseImmediateDef.
+Require Import compiler.UseImmediate.
 Import Utility.
 
 Section WithWordAndMem.
@@ -134,6 +134,11 @@ Section WithWordAndMem.
                 List.length (compile_ext_call posmap1 pos1 stackoffset c) =
                 List.length (compile_ext_call posmap2 pos2 stackoffset c)).
 
+    Definition is5BitImmediate(x: Z) : bool :=
+      andb (Z.leb 0 x) (Z.ltb x 32).
+    Definition is12BitImmediate(x: Z) : bool :=
+      andb (Z.leb (-2048) x) (Z.ltb x 2048).
+
     Definition locals_based_call_spec{Var Cmd: Type}{locals: map.map Var word}
                (Exec: string_keyed_map (list Var * list Var * Cmd) ->
                       Cmd -> trace -> mem -> locals -> MetricLog ->
@@ -164,6 +169,12 @@ Section WithWordAndMem.
       Valid := map.forall_values ParamsNoDup;
       Call := locals_based_call_spec FlatImp.exec;
     |}.
+
+    (* |                 *)
+    (* | UseImmediate    *)
+    (* V                 *)
+    (* FlatWithStrVars   *)
+
     (* |                 *)
     (* | RegAlloc        *)
     (* V                 *)
@@ -246,6 +257,46 @@ Section WithWordAndMem.
           eapply ListSet.In_list_union_l. eapply ListSet.In_list_union_l. assumption.
         + eapply @freshNameGenState_disjoint_fbody.
       - simpl. intros. fwd. eauto using map.getmany_of_list_extends.
+    Qed.
+
+    Lemma useimmediate_functions_NoDup: forall funs funs',
+        (forall f argnames retnames body,
+          map.get funs f = Some (argnames, retnames, body) -> NoDup argnames /\ NoDup retnames) ->
+        (useimmediate_functions is5BitImmediate is12BitImmediate) funs = Success funs' ->
+        forall f argnames retnames body,
+          map.get funs' f = Some (argnames, retnames, body) -> NoDup argnames /\ NoDup retnames.
+    Proof.
+      unfold useimmediate_functions. intros.
+      eapply map.try_map_values_bw in H0.
+      2: { eassumption.  }
+      unfold useimmediate_function in *.
+      fwd.
+      eapply H.
+      eassumption.
+    Qed.
+
+    Lemma useimmediate_correct: phase_correct FlatWithStrVars FlatWithStrVars (useimmediate_functions is5BitImmediate is12BitImmediate).
+    Proof.
+      unfold FlatWithStrVars.
+      split; cbn.
+      { unfold map.forall_values, ParamsNoDup. intros.
+        destruct v as  ((argnames & retnames) & body).
+        eapply useimmediate_functions_NoDup; try eassumption.
+        intros. specialize H0 with (1 := H2).
+        simpl in H0. assumption.
+      }
+
+      unfold locals_based_call_spec. intros. fwd.
+      pose proof H0 as GI.
+      unfold  useimmediate_functions in GI.
+      eapply map.try_map_values_fw in GI. 2: eassumption.
+      unfold useimmediate_function in GI. fwd.
+      eexists _, _, _, _. split. 1: eassumption. split. 1: eassumption.
+      intros.
+      eapply exec.weaken.
+      - eapply useImmediate_correct_aux.
+        all: eauto.
+      - eauto.
     Qed.
 
     Lemma regalloc_functions_NoDup: forall funs funs',
@@ -386,17 +437,19 @@ Section WithWordAndMem.
       string_keyed_map (list string * list string * cmd) ->
       result (list Instruction * string_keyed_map Z * Z) :=
       (compose_phases flatten_functions
+      (compose_phases (useimmediate_functions is5BitImmediate is12BitImmediate)
       (compose_phases regalloc_functions
       (compose_phases spill_functions
-                      (riscvPhase compile_ext_call)))).
+                      (riscvPhase compile_ext_call))))).
 
     Lemma composed_compiler_correct: phase_correct SrcLang RiscvLang composed_compile.
     Proof.
       unfold composed_compile.
       exact (compose_phases_correct flattening_correct
+            (compose_phases_correct useimmediate_correct
             (compose_phases_correct regalloc_correct
             (compose_phases_correct spilling_correct
-                                    riscv_phase_correct))).
+                                    riscv_phase_correct)))).
     Qed.
 
     Definition compile(funs: list (string * (list string * list string * cmd))):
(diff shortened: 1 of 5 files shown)


samuelgruetter (commented):
This looks great!

When studying the diff that you pasted, I noticed the following pattern on l532:

```
addi    x6, x0, 11
add     x10, x0, x6
```

and I was wondering why it was not optimized into

```
addi    x6, x0, 11
addi    x10, x0, 11
```

If you're interested in a little debugging/analysis exercise, stop reading now :wink: Otherwise, my solution/current understanding follows below:

I wanted to read the source code of all the lightbulb functions, so I inserted the following code in `End2EndLightbulb.v`:

```coq
Ltac unfold_all_in l x :=
  lazymatch l with
  | nil => idtac
  | cons (_, ?d) ?t => unfold d in x; unfold_all_in t x
  end.

Set Printing Depth 10000.

Goal True.
  pose (x := function_impls).
  unfold function_impls in x.
  lazymatch goal with
  | y := ?body |- _ => unfold_all_in body y
  end.
```

And then I searched for the constant `11`, quite a particular value, which happened to be unique, and appears in `lan9250_readword` (defined in `LAN9250.v`).
The relevant line of that function in that file is:

```
    unpack! ret, err = spi_xchg($0x0b);        require !err; (* FASTREAD *)
```

So we see that `0x0b` aka `11` is passed as a function argument. So this pattern was actually not yet present when `UseImmediate` ran, but only introduced by the spilling phase, which copies all function argument registers (`x6` in this example) into the argument registers (`x10`, `x11`, ... `x17`, but since this function takes only one argument, only `x10` is used). So if we wanted to optimize this, it would have to be some kind of constant propagation phase that runs after the spilling phase.

And a nitpick: commit messages don't need to be detailed, but maybe sliiiightly more descriptive than just "first commit", eg "wip calling UseImmediate in Pipeline", would be good.

And something else that I found a little surprising when looking at the diff: I was expecting that the `UseImmediate` phase without a dead code elimination phase would make the final output worse, because it introduces dead code, and I thought this would confuse the register allocator, because dead variables are never "read for the last time", and therefore it might think that they are actually live until the end of the function, and take up registers that then can't be used for other variables. But it turns out that this does not happen, because the register allocator is actually not as bad as I thought. In particular, on these lines:

https://github.com/mit-plv/bedrock2/blob/ef2d50031e88bbc92b2f64c7f1e9dbba9d63cdfc/compiler/src/compiler/RegAlloc.v#L67-L73

we can see that a `start_event` (variable is written) also inserts an end event (for a different reason than what I described above), but it has the effect that if this end event happens to be the last end event, the register allocator correctly understands that the live interval of that dead variable ends right there. So this means that even without a dead code elimination phase, we could merge your code that invokes the `UseImmediate` phase, and get code with exactly the same number of instructions.

So, to summarize: Looks great, no need to comment out your changes, and let me know if you need help with the incomplete proofs.

compiler/src/compiler/Pipeline.v
```diff
@@ -134,6 +135,12 @@ Section WithWordAndMem.
                 List.length (compile_ext_call posmap1 pos1 stackoffset c) =
                 List.length (compile_ext_call posmap2 pos2 stackoffset c)).
 
+
+    Definition is5BitImmediate : Z -> bool :=
+      fun x => (andb (Z.leb 0 x) (Z.ltb x 32)).
+    Definition is12BitImmediate : Z -> bool :=
+      fun x => (andb (Z.leb 0 x) (Z.ltb x 2048)).
```
samuelgruetter, line 142: I think most (or all?) 12-bit immediates are signed, ie they should be between `-2^11` and `2^11`.
For instance, the check for instructions in the I format is here:
https://github.com/mit-plv/riscv-coq/blob/55c9cc88a6734550bdd913bfdf482d498c6354e6/src/riscv/Utility/Encode.v#L311-L316
But you should check in the spec for each instruction which format it uses (probably most are in the I format), and also check the 5-bit ones: are they only used for bitshifts, and are they unsigned?
It would be good to add a test that explores some of these boundaries, and proves `Forall (fun inst => verify inst RV32I)) compiled_instructions`, because we don't have any proof that the compiler will always output valid instructions, it's only tested for each program after compiling it, so it should be tested a bit in advance. I'd do this in a test-driven manner: First, write the test (ie a proof that can't be proven), commit it, then in the next commit, fix `is12BitImmediate` (if needed).

compiler/src/compiler/Pipeline.v
```diff
@@ -248,6 +260,50 @@ Section WithWordAndMem.
...
+      eapply map.try_map_values_fw in GI. 2: eassumption.
+      unfold useimmediate_function in GI. fwd.
+      eexists _, _, _, _. split. 1: eassumption. split. 1: eassumption.
+      intros.
+      eapply exec.weaken.
+      - eapply useImmediate_correct_aux.
+        all: eauto.
+      - eauto.
+
+
```
samuelgruetter, line 304: Formatting nitpick: Before finalizing this, please make sure that your use of whitespace makes sense (for a vague definition of "makes sense" :wink:). Two empty lines before a `Qed` doesn't exactly make sense to me... Also read your diff (`git diff`) and avoid introducing or deleting empty lines at random unmodified locations

compiler/src/compilerExamples/immediateExample.v
```diff
@@ -92,6 +77,10 @@ Definition imm_asm: list Instruction.
   end.
 Defined.
 
+Goal Forall (fun inst => verify inst RV32I) imm_asm.
+  all: repeat constructor; simpl.
+Admitted.
```
0adb, line 82: I'm currently working on examples for checking that the assembly produced is valid, and I'm currently stuck on goals that look like image attached , where one side is a Z and the other is a constant.
image
I'm unsure how to solve these goals without repeatedly using `replace {constant_name} with {value}` for each named constant, followed by `blia` and `auto`, and would like to ask if there is a way to automate this.

samuelgruetter: Ah I see, ok, so you could probably do some ad-hoc hacks like `cbv; intros; try congruence`, etc, but actually, the better way of doing this is to run a Gallina checker rather than proving everything using Ltac.
That is, instead of proving `Forall (fun i => verify i RV32I)`, prove `bverify.validInstructions RV32I`, because this one should evaluate to `true` just by calling `vm_compute`.
I did such a change recently in here https://github.com/mit-plv/bedrock2/commit/11e77912b166bc3ed14c00829d9f7a4af2e58ebe
Let me know if you need more details

compiler/src/compilerExamples/immediateExample.v
```diff
@@ -54,50 +54,86 @@ Notation RiscvMachine := MetricRiscvMachine.
...
 
-Lemma f_equal2: forall {A B: Type} {f1 f2: A -> B} {a1 a2: A},
-    f1 = f2 -> a1 = a2 -> f1 a1 = f2 a2.
-Proof. intros. congruence. Qed.
+Definition imm_ex5 : (string * (list string * list string * cmd )) := test_imm_op bopname.srs 0 31.
 
-Lemma f_equal3: forall {A B C: Type} {f1 f2: A -> B -> C} {a1 a2: A} {b1 b2: B},
-    f1 = f2 -> a1 = a2 -> b1 = b2 -> f1 a1 b1 = f2 a2 b2.
-Proof. intros. congruence. Qed.
+Definition imm_ex6 : (string * (list string * list string * cmd )) := test_imm_op bopname.slu 0 31.
```
0adb, line 78: Would like to ask if these tests seem broad enough in terms of coverage. I think I was able to get the proof that the assembly produced by these to be valid to go through later in the file.

samuelgruetter: Coverage looks reasonable 👍 
Just note that vm_cast_no_check delays checking until the Qed, so if you then do Abort instead of Qed, you're not checking anything! On these small examples, you don't need to use the unsafe vm_cast_no_check, just use tactics that you know. And maybe also Qed it, the reason I sometimes do Abort instead is to avoid registering spurious theorems (that will show up in Search later), but this example will probably never be used by other files, so it doesn't matter how many theorems it registers.

compiler/src/compilerExamples/immediateExample.v
```diff
@@ -54,41 +48,64 @@ Notation RiscvMachine := MetricRiscvMachine.
...
+      match h_res with
+      | Success (x, _, _) =>
+          let t_res := multi_compile t in
+          match t_res with
+          | Success t => Success (x :: t)
+          | _ => t_res
+          end
+      | Failure x => Failure x
+      end
+end.
```
samuelgruetter, line 95: Applying a `result`-returning function to all elements of a list looks like something quite generic. Are you sure there's not already a library function that does this for you? For instance, if you do `Search (list (result ?A) -> result (list ?A))`, it returns `List.all_success`, could you make use of this one?

compiler/src/compilerExamples/immediateExample.v
samuelgruetter, line 1: This file is now quite a good test, but it still looks a bit like the result of copying another test file, ie it still contains a lot of unrelated cruft. For instance, does `compile_ext_call` really need to be so long? Can't it just always return the empty list? And what is `main_stackalloc` good for? Are you interested in testing anything related to stack allocation here? And `Notation RiscvMachine := MetricRiscvMachine.` seems unneded as well.

samuelgruetter (changes requested):

compiler/src/compilerExamples/immediateExample.v
```diff
@@ -1,107 +1,84 @@
...
-Definition imm_asm: list Instruction.
-  let r := eval cbv in (compile compile_ext_call allFuns) in set (res := r).
+Definition multi_compile funs_lists :=
+  List.all_success
+    (List.map
+       (fun x =>
+          match x with
+          | Success (insts, _, _) => Success insts
+          | Failure e => Failure e
+          end)
```
0adb, line 52: Would like to ask if there is a simpler way to write this. From `coqutil/Datatypes/Result.v` 
coqutil-result-def
there seems to be notation for doing something like
image
but trying the latter gives an error about `((insts, _, _))` should be an identifier instead.

samuelgruetter: Could it be that you just forgot to put a single quote before the opening parenthesis of the pattern?

0adb: Yep, I think that adding the single quote made it work.

PR #344, opened 2023-04-06
Preliminary work on expression compiler

psvenk:
This is not ready to merge (and some conflicts still need to be resolved, anyway), but I figured that I would open a draft pull request for now after having sat on this code for a while.

### DIFF
--- a/PyLevelLang/src/PyLevelLang/Compile.v
+++ b/PyLevelLang/src/PyLevelLang/Compile.v
@@ -139,7 +139,44 @@ Section WithMap.
     eval_expr_old m l e = Some w ->
     eval_expr_old m l' e = Some w.
   Proof.
-  Admitted.
+    induction e; intros m l l' w H Hl.
+    - (* Syntax.expr.literal *)
+      exact Hl.
+    - (* Syntax.expr.var *)
+      simpl in Hl. simpl. unfold map.extends in H.
+      now apply H.
+    - (* Syntax.expr.inlinetable *)
+      simpl in Hl. simpl.
+      specialize IHe with (m := m) (1 := H).
+      destruct (eval_expr_old m l e); try easy.
+      specialize IHe with (w := r) (1 := eq_refl).
+      now rewrite IHe.
+    - (* Syntax.expr.load *)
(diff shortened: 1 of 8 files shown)


samuelgruetter:
I didn't manage to isolate the `Qed` typechecking failure bug, but at least, I found a workaround: https://github.com/mit-plv/bedrock2/commit/12a789ec7f543dec62e0bc1a6eece6569cf26c9a
It just consists closing the current `Section` and starting a new `Section` whenever the bug occurs.
But I don't consider this very nice code.
So I would suggest the following alternative, which is good practice anyways: Split `Optimize.v` into two files, one for the definitions of the optimization functions (eg `OptimizeDefs.v`) and one for the proofs (eg `OptimizeProofs.v`).
/cc @andres-erbsen

samuelgruetter:
> Ha this sure sounds like one of the tactics involved (probably `dependent induction`) does the wrong thing when referencing definitions that depend on section variables. Perhaps it needlessly generalizes the context but still uses the env reference that depends on pre-generalization context.

Yes, that's exactly what happens. *Why* it happens, that's a different question: I tried to reproduce it from scratch in an empty file, but the variations I tried all worked correctly, and my time budget is exceeded :sweat_smile:  Next step would be to isolate it by simplifying the original file manually, or by throwing the bug minimizer at it, if you want to give it a try :wink:

andres-erbsen:
Reported at https://github.com/coq/coq/issues/17555

samuelgruetter:
> Reported at [coq/coq#17555](https://github.com/coq/coq/issues/17555)

Cool :tada: While minimizing, did you find a better workaround than what I suggested above?

andres-erbsen:
No, and i'm not really holding my breadth that there would be one.

psvenk:
> So I would suggest the following alternative, which is good practice anyways: Split `Optimize.v` into two files, one for the definitions of the optimization functions (eg `OptimizeDefs.v`) and one for the proofs (eg `OptimizeProofs.v`).

This did not work because the issue is dependence between the proofs and not dependence of the proofs on the definitions. So, I applied 12a789e with the Coq bug report linked in comments.

PR #347, opened 2023-04-06
Initial progress on encoding SQL queries and related optimizations

hulsemohit:
Queries.v has two ways to encode SQL queries in the language (with some helper functions).
I've also added a few optimizations to fuse fold and flatmap.
I needed to add a few lemmas about decidable equality of string/bool/ascii, which I'm told should probably eventually be moved to Coqutil (they're currently still in Optimize.v).

### DIFF
--- a/PyLevelLang/src/PyLevelLang/Queries.v
+++ b/PyLevelLang/src/PyLevelLang/Queries.v
@@ -1,0 +1,114 @@
+Require Import PyLevelLang.Language.
+Require Import PyLevelLang.Elaborate.
+Require Import PyLevelLang.Notations.
+Require Import PyLevelLang.Interpret.
+Require Import PyLevelLang.SamplePrograms.
+Require Import PyLevelLang.Optimize.
+Require Import coqutil.Map.Interface coqutil.Map.SortedListString coqutil.Map.Properties.
+Require Import coqutil.Datatypes.Result.
+Require Import Coq.Lists.List.
+
+Local Open Scope Z_scope.
+Local Open Scope string_scope.
+Local Open Scope list_scope.
+
+
+Section Queries_Section.
+  Instance tenv : map.map string (type * bool) := SortedListString.map _.
+  Instance tenv_ok : map.ok tenv := SortedListString.ok _.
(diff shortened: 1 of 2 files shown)


PyLevelLang/src/PyLevelLang/Queries.v
```diff
@@ -0,0 +1,114 @@
...
+
+  Definition test := PEProj (PERecord (("a", PEAtom (PAInt 1)) :: ("b", PEAtom (PAInt 2)) :: nil)) "a".
+
+  Fixpoint pexpr_list (t : type) (l : list (pexpr)) : pexpr :=
+    match l with
+    | nil => PEAtom (PANil t)
+    | x :: xs => PEBinop POCons x (pexpr_list t xs)
+    end.
+
+  Fixpoint as_pexpr {t : type} (v : interp_type t) : pexpr :=
```
DIJamner, line 69: This seems like an odd place for this function. If it doesn't already exist, we should move it somewhere else, probably to where `pexpr` is defined.

PyLevelLang/src/PyLevelLang/Queries.v
```diff
@@ -0,0 +1,114 @@
...
+    match t return interp_type t -> pexpr with
+    | TInt => fun v => PEAtom (PAInt v)
+    | TBool => fun v => PEAtom (PABool v)
+    | TString => fun v => PEAtom (PAString v)
+    | TPair _ _ _  => fun v => PEBinop POPair (as_pexpr (fst v)) (as_pexpr (snd v))
+    | TEmpty => fun v => PERecord nil
+    | TList t => fun v => pexpr_list t (map as_pexpr v)
+    end v.
+
+  Fixpoint zip {A B} (l1 : list A) (l2 : list B) : list (A * B) :=
```
DIJamner, line 79: We have `coqutil.Datatypes.List.zip`, which when passed `pair` as its first argument, is equal to this function.

DIJamner (approved):
While I'd like to make the improvements I've mentioned, I'm fine merging as-is and fixing them later.

PR #355, opened 2023-07-03
read Ltac2 integer bits using bitwise right shift

robertzhidealx:
Now that Ltac2 [supports](https://github.com/coq/coq/pull/15637) integer operations like the bitwise right shift (`Int.lsr`), we can use it directly for converting an Ltac2 integer into a bool list representation of its bits, obviating the original workaround.

Compiles in Coq version 8.17.0.


robertzhidealx:
Please squash and merge if approved. Thanks!

samuelgruetter:
nice, thanks :+1:

PR #356, opened 2023-07-05
Stop using auto with * in intuition

SkySkimmer:
Needs https://github.com/mit-plv/kami/pull/35


samuelgruetter:
See https://github.com/mit-plv/bedrock2/commit/9320f7c9b92b0daffd57c4251302334556299197

PR #359, opened 2023-07-21
Support for let bindings in fiat2->bedrock2 expression compiler

psvenk:
This still contains some `admit`s relating to proving that names are not reused, which turned out to be harder than expected. I plan on resolving these by creating a new intermediate language that does not allow shadowing (we need to add an intermediate language for notating memory lifetimes anyway), so I figured that I would open a pull request for this separately from those more substantial changes.

### DIFF
--- a/PyLevelLang/Makefile
+++ b/PyLevelLang/Makefile
@@ -13,7 +13,7 @@ DEPS_DIR ?= ../deps
 # Note: make does not interpret "\n", and this is intended
 DEPFLAGS_COQUTIL_NL=-Q $(DEPS_DIR)/coqutil/src/coqutil coqutil\n
 DEPFLAGS_NL=
-CURFLAGS_NL=-Q ../bedrock2/src/bedrock2 bedrock2\n-Q src/PyLevelLang PyLevelLang\n
+CURFLAGS_NL=-Q ../bedrock2/src/bedrock2 bedrock2\n-Q ../compiler/src/compiler compiler\n-Q src/PyLevelLang PyLevelLang\n
 
 EXTERNAL_DEPENDENCIES?=
 EXTERNAL_COQUTIL?=
--- a/PyLevelLang/src/PyLevelLang/Compile.v
+++ b/PyLevelLang/src/PyLevelLang/Compile.v
@@ -7,6 +7,7 @@ Require Import coqutil.Datatypes.Result.
 Import ResultMonadNotations.
 Require Import coqutil.Map.Interface coqutil.Map.SortedListString coqutil.Map.MapEauto.
 Require Import coqutil.Byte coqutil.Word.Interface coqutil.Word.Bitwidth coqutil.Word.Properties.
+Require Import compiler.NameGen.
 
 Require Import coqutil.Tactics.Tactics.
 Require Import coqutil.Tactics.fwd.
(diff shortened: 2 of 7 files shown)


PyLevelLang/src/PyLevelLang/Compile.v
```diff
@@ -76,32 +77,64 @@ Definition compile_binop {t1 t2 t3 : type} (o : binop t1 t2 t3) :
...
-  | EBinop o e1 e2 =>
-      f <- compile_binop o;;
-      '(c1, e1') <- compile_expr e1;;
-      '(c2, e2') <- compile_expr e2;;
-      Success (Syntax.cmd.seq c1 c2, f e1' e2')
-  | _ => error:("unimplemented")
-  end.
+Section Map.
+  Context {key value : Type} {map : map.map key value} {map_ok : map.ok map}.
+  Definition map_values (m : map) : list value := map.fold (fun acc k v => cons v acc) nil m.
```
samuelgruetter, line 82: If you really need this function, I would define it in terms of `map.tuples`, `List.map`, and `snd` (avoid `map.fold` whenever you can because it tends to require tricky inductive hypotheses).

PyLevelLang/src/PyLevelLang/Compile.v
```diff
@@ -76,32 +77,64 @@ Definition compile_binop {t1 t2 t3 : type} (o : binop t1 t2 t3) :
...
+
+  Fixpoint compile_expr {t : type} (e : expr t) (nm : namemap) (st : NGstate) :
+    result (Syntax.cmd * Syntax.expr * namemap * NGstate) :=
+    match e with
+    | EVar _ x | ELoc _ x =>
+        match map.get nm x with
+        | Some x' =>
+            Success (Syntax.cmd.skip, Syntax.expr.var x', nm, st)
+        | None =>
+            error:("variable x not mapped in target language (unreachable)")
```
samuelgruetter, line 99:
```suggestion
            error:("variable" x "not mapped in target language (unreachable)")
```

PyLevelLang/src/PyLevelLang/Compile.v
```diff
@@ -76,32 +77,64 @@ Definition compile_binop {t1 t2 t3 : type} (o : binop t1 t2 t3) :
...
+        '(c1, e1', _, _) <- compile_expr e1 nm st;;
+        Success (c1, f e1', nm, st)
+    | EBinop o e1 e2 =>
+        f <- compile_binop o;;
+        '(c1, e1', _, _) <- compile_expr e1 nm st;;
+        '(c2, e2', _, _) <- compile_expr e2 nm st;;
+        Success (Syntax.cmd.seq c1 c2, f e1' e2', nm, st)
+    | ELet x e1 e2 =>
+        '(c1, e1', _, _) <- compile_expr e1 nm st;;
+        '(c2, e2', _, _) <- compile_expr e2 nm st;;
```
samuelgruetter, line 115: I think here (and in the other recursive calls too) you should properly thread through the `nm` and the `st`, because the target language doesn't support shadowing. I.e. don't just ignore what the recursive calls return, but pass it to the next recursive call, and use what the second recursive call returned in the `map.put`.
Or, alternatively, if it's ok if evaluation of different subexpressions overwrites previously used variables, that would mean you can really ignore the `nm` and `st` return values and don't actually need to return them? At the moment, it seems that you don't really read them anywhere.

PyLevelLang/src/PyLevelLang/Compile.v
```diff
@@ -113,19 +146,130 @@ Section WithMap.
...
+  Definition locals_relation (lo : locals) (l : locals') (nm : namemap) : Prop :=
+    map.forall_keys (fun x =>
+    match map.get lo x, map.get nm x with
+    | Some (existT _ _ val), Some x' =>
+        match map.get l x' with
+        | Some val' => value_relation val val'
+        | None => False
+        end
+    | _, _ => True
+    end) lo.
```
samuelgruetter, line 161: This looks a bit more complicated than it needs to be. Further down, in `locals_relation_get`, you prove that
```coq
    locals_relation lo l nm
```
implies something more usable:
```coq
    forall (x x' : string) (t : type) (s : {t & interp_type t}),
    map.get lo x = Some s ->
    map.get nm x = Some x' -> exists (w : word),
    map.get l x' = Some w
```
This could actually be the body of the definition, and you won't need a `locals_relation_get` lemma any more, and I would also expect that `locals_relation_put` will become simpler.

PyLevelLang/src/PyLevelLang/Compile.v
```diff
@@ -113,19 +146,130 @@ Section WithMap.
...
-  (*   end v. *)
+  Definition tenv_relation (G : tenv) (lo : locals) : Prop :=
+    map.forall_keys (fun key =>
+    match map.get G key with
+    | Some (t, _) => match map.get lo key with
+                     | Some (existT _ t' _) => t = t'
+                     | None => False
+                     end
+    | None => True
+    end) G.
```
samuelgruetter, line 224: Same simplification opportunity as for `locals_relation`

PyLevelLang/src/PyLevelLang/Compile.v
```diff
@@ -134,78 +278,149 @@ Section WithMap.
...
-    wf map.empty e ->
-    compile_expr e = Success (c, e') -> forall tr m l mc,
+  Lemma compile_correct :
+    forall {t} (e : expr t) (c : Syntax.cmd) (e' : Syntax.expr)
+    (G : tenv) (lo : locals) (nm nm' : namemap) (st st' : NGstate),
+    tenv_relation G lo ->
+    wf G e ->
+    compile_expr e nm st = Success (c, e', nm', st') -> forall tr m l mc,
+    locals_relation lo l nm ->
+    locals_relation lo l nm' /\
```
samuelgruetter, line 375: I think the left hand side of this `/\` could be split off into two simple separate lemmas:
* `compile_expr e nm st = Success (c, e', nm', st') -> map.extends nm' nm`
* `map.extends nm' nm -> locals_relation lo l nm -> locals_relation lo l nm'`

samuelgruetter (commented):
Looks great :+1: I added some comments on specific lines.

Regarding
> creating a new intermediate language that does not allow shadowing 

I think the whole point of using a namemap and a fresh name generator is to be able to deal with shadowing right here in this phase. If you add a separate deshadowing phase anyways, you don't need the namemap and fresh name gen here and can just use the same names in the source and target language here.

PyLevelLang/Makefile
```diff
@@ -13,7 +13,7 @@ DEPS_DIR ?= ../deps
 # Note: make does not interpret "\n", and this is intended
 DEPFLAGS_COQUTIL_NL=-Q $(DEPS_DIR)/coqutil/src/coqutil coqutil\n
 DEPFLAGS_NL=
-CURFLAGS_NL=-Q ../bedrock2/src/bedrock2 bedrock2\n-Q src/PyLevelLang PyLevelLang\n
+CURFLAGS_NL=-Q ../bedrock2/src/bedrock2 bedrock2\n-Q ../compiler/src/compiler compiler\n-Q src/PyLevelLang PyLevelLang\n
```
samuelgruetter, line 16: If you do this, you should also say in the toplevel makefile that `PyLevelLang` now depends on `compiler`

psvenk:
> > creating a new intermediate language that does not allow shadowing
> 
> I think the whole point of using a namemap and a fresh name generator is to be able to deal with shadowing right here in this phase. If you add a separate deshadowing phase anyways, you don't need the namemap and fresh name gen here and can just use the same names in the source and target language here.

Yes; what I meant is that the preprocessing step would take care of deshadowing (using a namemap and fresh name generator) and `Compile.v` would just use the same names in the source and target languages. I realized that this can be done in the existing elaborate phase (by making `wf` require deshadowing), so I'll take care of that in this pull request itself.

samuelgruetter:
Let's make sure we distinguish between *shadowing* (eg `let x = 1 in (let x = 2 in x)`) and *reusing* (eg `(let x = 1 in x) + (let x = 2 in x)`). I think a source program with reusing should not be a problem.
As a next step, I would recommend thinking more about the signature of `compile_expr` (see also [this comment](https://github.com/mit-plv/bedrock2/pull/359/files#r1271140802)). Once you have a clear understanding how much you want or don't want to thread name maps and namgen states through `compile_expr`, the rest should become clearer too. And an additional idea for that function's signature would be to give it as an argument the name of the variable to which you want the result to be assigned. This could be important to get compilation of expressions to be modular enough: After the source statement `(let x = a in b)`, `x` is not in scope any more, so expecting that it is in scope after the compiled statement corresponding to that source statement would seem to introduce weird constraints (not sure if you were actually trying this). Or said differently, consider this example: `(let x = 1 in (let y = 2 in (let z = x + y in z)))`. Its result value is stored in `z`, but expecting that `z` is still available after that statment seems wrong, because z's scope ends after the first `)`.

PR #386, opened 2023-11-08
Use List.Forall2_nil instead of List.Forall2_refl

SkySkimmer:
Forall2_refl is deprecated due to its type not matching its name (it's only about nil), Forall2_nil having the same type.


samuelgruetter:
Looks like this is an easy one to merge, for once :wink: 
Thanks for all your work to keep Bedrock2 compatible :+1:

Issue #388, opened 2023-11-19
install target is unusable when sudo doesn't have access to coqc


JasonGross:
`make install` should not be invoking bytedump test, or at the very least the bytedump test should respect COQBIN
```
 COQFLAGS="-Q src/bedrock2 bedrock2 -Q src/bedrock2Examples bedrock2Examples -Q /github/workspace/rupicola/bedrock2/deps/coqutil/src/coqutil coqutil " ../etc/bytedump.py bedrock2.PrintListByte.allBytes > special/BytedumpTest.out.tmp
Traceback (most recent call last):
  File "/github/workspace/rupicola/bedrock2/bedrock2/../etc/bytedump.py", line 16, in <module>
    p = subprocess.run(["coqtop", "-q", "-quiet"] + COQFLAGS, stdout=subprocess.PIPE, stderr=subprocess.PIPE, input=os.fsencode(f"""
  File "/usr/lib/python3.9/subprocess.py", line 505, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.9/subprocess.py", line 1823, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'coqtop'
Error: make[2]: *** [Makefile:64: special/BytedumpTest.out] Error 1
```
https://github.com/mit-plv/fiat-crypto/actions/runs/6921348406/job/18826839337?pr=1733#step:18:3326

samuelgruetter:
I'm trying to understand why these 4 dependencies of the `test` target are always run:

https://github.com/mit-plv/bedrock2/blob/463e132a5b8b8e841ffbcca73b37619dd9505a9f/bedrock2/Makefile#L49

For `typecheckExprToCString`, `testStackLoop`, `testStackNondet`, I think one reason is that they don't create a file with that name, so running these tests does not leave any trace of when it has been run last, so `make` has to rerun it each time. An easy fix for these would probably be to have the tests create a `.ok` file.
So I tried the following patch in `bedrock2/Makefile`:

```diff
        $(BYTEDUMP) bedrock2.ToCStringStackallocLoopTest.main_cbytes > special/stackloop.c
 special/stackloop: special/stackloop.c
        $(CC) -O0 special/stackloop.c -o special/stackloop
[...]

PR #391, opened 2023-11-22
Don't rerun tests in bedrock2/Makefile each time

samuelgruetter:
Does this fix https://github.com/mit-plv/bedrock2/issues/388, @JasonGross?


JasonGross:
Can we merge and see if it works in fiat-crypto's CI?

JasonGross:
[Seems to work](https://github.com/mit-plv/fiat-crypto/actions/runs/6963911933/job/18950293249?pr=1740), thanks!

PR #396, opened 2024-01-05
sepapps safe to cancel

vfukala:
In LiveVerif: Making `predicates_safe_to_cancel` succeed if both the hypothesis and the conclusion is a `sepapps` and all the corresponding constituent predicates are safe to cancel (checked recursively). Adding a test for this (which would previously fail).


samuelgruetter:
Looks great, thanks!

PR #397, opened 2024-01-05
Try assumption when checking if purified hyp already derivable

vfukala:
Instead of just running `xlia`, checking if the purified version of a hyp can already be derived now also splits all `and`s and tries the `assumption` tactic.

### DIFF
--- a/bedrock2/src/bedrock2/PurifyHeapletwise.v
+++ b/bedrock2/src/bedrock2/PurifyHeapletwise.v
@@ -9,8 +9,11 @@ Require Import coqutil.Tactics.ident_ops.
 Require Import bedrock2.bottom_up_simpl.
 Require Import bedrock2.unzify.
 
+Ltac split_and_solve_trivial :=
+  repeat match goal with | |- _ /\ _ => split end; try assumption.
+
 Ltac fail_if_too_trivial t :=
-  assert_fails (idtac; assert t by xlia zchecker).
+  assert_fails (idtac; assert t by (split_and_solve_trivial; xlia zchecker)).
 
 Ltac puri_simpli_zify_hyp fastMode h t :=
   let pure := purified_hyp h t in
@@ -48,7 +51,8 @@ Inductive derivability_test_marker: Prop := mk_derivability_test_marker.
 
 Ltac clear_pure_hyp_if_derivable h tp :=
   tryif ident_starts_with __pure_ h then
-    try (clear h; assert_succeeds (idtac; assert tp by (zify_goal; xlia zchecker)))
+    try (clear h; assert_succeeds (idtac; assert tp by
+      (split_and_solve_trivial; zify_goal; xlia zchecker)))
   else idtac.
 
 Ltac clear_upto_marker marker :=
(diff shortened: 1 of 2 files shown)


bedrock2/src/bedrock2/PurifyHeapletwise.v
```diff
@@ -9,8 +9,11 @@ Require Import coqutil.Tactics.ident_ops.
 Require Import bedrock2.bottom_up_simpl.
 Require Import bedrock2.unzify.
 
+Ltac split_and_solve_trivial :=
+  repeat match goal with | |- _ /\ _ => split end; try assumption.
```
samuelgruetter, line 13: `assumption` runs full unification that can arbitrarily unfold terms, so I'd like to keep it out of the library (and only use it in examples where it turns out to not hurt). Would the following work instead?
```suggestion
  repeat match goal with
  | |- _ /\ _ => split
  | H: ?P |- ?P => exact H
  end.
```

bedrock2/src/bedrock2/PurifyHeapletwise.v
```diff
@@ -9,8 +9,11 @@ Require Import coqutil.Tactics.ident_ops.
 Require Import bedrock2.bottom_up_simpl.
 Require Import bedrock2.unzify.
 
+Ltac split_and_solve_trivial :=
```
samuelgruetter, line 12: Having `_solve_` in the name is a bit misleading, because it suggests that it behaves like the `solve` tactic, ie. fully solves the goal or fails, but the way this tactic is implemented, it can also leave open some goals. Is this intended? If so, I'd remove `_solve_` from the name. Or else, I'd add `solve` around it.

samuelgruetter (changes requested):

bedrock2/src/bedrock2/PurifyHeapletwise.v
JasonGross: Nonlinear patterns also use unification (at least to some extent) when some copies of the metavariable are in the goal and other copies are in hypotheses.  (If all copies are in the goal or all copies are in hypotheses, syntactic matching is done instead). (Unless @ppedrot has "fixed" this behavior?)

Consider instead
```suggestion
  repeat match goal with
  | |- _ /\ _ => split
  | H: ?P1 |- ?P2 => constr_eq_nounivs P1 P2; exact H
  end.
```

ppedrot: This behaviour was not changed, and is unlikely to change, because *backwards compatibility*...

ppedrot: Note also that there is still a slight difference between assumption and non-linear matching w.r.t. universe unification. The first one can unify universes, while the latter does not and relies on syntactic matching of universes.

While I'm at it and to make things actually technically correct, none of these are performing *term unification* as suggested by @samuelgruetter. Rather, they're actually checking for *term conversion* (potentially up to universe unification). Conversion can indeed arbitrarily unfold stuff but it's in general faster than full-blown unification which may instantiate evars and try weird heuristics.

samuelgruetter: Interesting, thanks @JasonGross and @ppedrot for the explanations!

PR #399, opened 2024-01-24
Crit-bit Trees: Lookup, Insert, Delete

vfukala:
This PR focuses on the verified implementation of crit-bit trees, which is part of the repository as a LiveVerif example.

This PR is a partial rewrite of the previous implementation. The new implementation should be more maintainable and extensible. Further, this PR
* adds an implementation of a delete function for crit-bit trees, and
* doesn't `admit` any lemmas used in the verification of crit-bit trees. This is in contrast to the pre-PR code, which does admit some.


samuelgruetter:
This looks great :tada: 
My only concern is why `cbt_insert` has a suspicious `* fun _ => True }> m'` in the postcondition. This line was already merged, and I think I already asked this question. IIRC, it was for the case where you need to allocate two new nodes, and allocating the first one succeeds, but allocating the second one fails? Then, you'd want to free the first, but you can't because to free something, you need an `allocator`, but you only have an `allocator_failed_below 12`. Does that sound right?
This is not specific to critbit trees, but should be solved in the malloc library -- though I don't quite know how yet...

vfukala:
> This looks great 🎉 My only concern is why `cbt_insert` has a suspicious `* fun _ => True }> m'` in the postcondition. This line was already merged, and I think I already asked this question. IIRC, it was for the case where you need to allocate two new nodes, and allocating the first one succeeds, but allocating the second one fails? Then, you'd want to free the first, but you can't because to free something, you need an `allocator`, but you only have an `allocator_failed_below 12`. Does that sound right? This is not specific to critbit trees, but should be solved in the malloc library -- though I don't quite know how yet...

Thanks for looking at this and yes, all that sounds right!

PR #400, opened 2024-01-29
use `flat_map_constant_length` from stdlib instead of equivalent `length_flat_map` from coqutil

andres-erbsen:


andres-erbsen:
:man_facepalming:  ubuntu https://repology.org/project/coq/badges https://github.com/mit-plv/bedrock2/pull/401

PR #412, opened 2024-04-06
only_mmio_satisfying

andres-erbsen:
please fast-forward


samuelgruetter:
I'm ok with these changes, but I don't think it's possible to fast-forward, because the parent of your commit ("only_mmio_satisfying") is "paperify lan9250_writepacket" (from Thu), but a little later on Thu, I already pushed another commit ("bump coqutil") to bedrock2 master. I'll stay out of the way and let you merge or fast-forward in whatever way works best for you.

PR #418, opened 2024-04-14
[CI] Move update-tested into coq.yml

JasonGross:
This version will only work if there's only one .yml file of checks we want to run, but it updates tested more quickly when there are pushes, and doesn't require keeping the list of Coq versions up to date in multiple places.

On top of #417


samuelgruetter:
I'm not so familiar with Coq's CI but if it indeed does not reference the `tested` branch any more, we can just delete `update-tested.yml` without moving its contents to `coq.yml` -- the fewer moving parts, the better!

JasonGross:
Sure, it can just be deleted instead

PR #425, opened 2024-06-21
compiler metrics

tckmn:
Some comments on the code as I'm looking the whole thing over now:

* I have only just now realized the ipow example no longer compiles, because I factored out lots of hardcoded constants into functions like `cost_op`. This should be extremely easy to fix at the tactic automation level (but I'm making this pull request now in the interest of being able to see the code sooner).

* In the previous code, the large constant bound on function calls was added twice in the semantics, once in the `exec fbody` judgment and once in the postcondition. I don't really understand why this is, but I kept them both (as `cost_call_internal` and `cost_call_external`). As far as I can tell, only one of these is needed, and putting it in either place works just as well.


compiler/src/compiler/FlatToRiscvFunctions.v
```diff
@@ -1950,6 +1988,20 @@ Section Proofs.
...
+                 end;
+          repeat match goal with
+                 | H : valid_FlatImp_var _ |- _ => apply valid_FlatImp_var_isRegZ in H; rewrite H in *
+                 end;
+          try MetricsToRiscv.solve_MetricLog.
+          destruct H6p0.
+          repeat match goal with
+                 | H : valid_FlatImp_var _ |- _ => apply valid_FlatImp_var_isRegZ in H; rewrite H in *
+                 end.
+          MetricsToRiscv.solve_MetricLog.
```
samuelgruetter, line 2004: This snippet is repeated quite a few times, could you factor it out into an Ltac, and also match on the shape of the hypothesis instead of hardcoding the name H6p0? And maybe you could also integrate it into run1done?

compiler/src/compiler/FlatToRiscvMetric.v
```diff
@@ -1,3 +1,6 @@
+(*
+TODO why did pratap delete this file?
```
samuelgruetter, line 2: There was a time when FlatToRiscvMetric contained a proof with metrics for everything except SCall and SInteract, and FlatToRiscvFunctions didn't do metrics, but treated the SCall and SInteract cases. Back then, we wanted to keep both files around. But now that FlatToRiscvMetric proves everything, FlatToRiscvMetric is not needed any more and can indeed be deleted.

compiler/src/compiler/LowerPipeline.v
```diff
@@ -676,7 +691,17 @@ Section LowerPipeline.
...
+        apply raise_metrics_ineq in H10p3.
+        destruct (getMetrics final).
+        destruct (getMetrics initial).
+        unfold raiseMetrics, lowerMetrics in *.
+        unfold_MetricLog.
+        simpl_MetricLog.
+        simpl in *.
+        inversion H2.
+        solve_MetricLog.
+        (* XXX ???? *)
```
samuelgruetter, line 704: What does `(* XXX ???? *)` mean here?

compiler/src/compiler/Pipeline.v
```diff
@@ -147,14 +159,29 @@ Section WithWordAndMem.
...
-               (t: trace)(m: mem)(argvals: list word)
-               (post: trace -> mem -> list word -> Prop): Prop :=
+               (t: trace)(m: mem)(argvals: list word)(mc: MetricLog)
+               (post: trace -> mem -> list word -> MetricLog -> Prop): Prop :=
+      exists argnames retnames fbody l,
+        map.get e f = Some (argnames, retnames, fbody) /\
+        map.of_list_zip argnames argvals = Some l /\
+        Exec e fbody t m l (cost_call_internal PreSpill mc) (fun t' m' l' mc' =>
+                       exists retvals, map.getmany_of_list l' retnames = Some retvals /\
+                                       post t' m' retvals mc').
```
samuelgruetter, line 169: The only difference between `locals_based_call_spec` and `locals_based_call_spec_spilled` is that the former wraps `mc` in `(cost_call_internal PreSpill mc)` before passing it to `Exec`. Can you get rid of the duplicate cost_call invocation that you mentioned in the PR, and then also get rid of the duplication here?

compiler/src/compiler/Pipeline.v
```diff
@@ -575,11 +609,14 @@ Section WithWordAndMem.
             word.unsigned ret_addr mod 4 = 0 ->
             arg_regs_contain initial.(getRegs) argvals ->
             initial.(getLog) = t ->
+            raiseMetrics (cost_SCall_L initial.(getMetrics)) = mcL ->
```
samuelgruetter, line 612: Can you get rid of the `cost_SCall_L` and also inline this equation?

compiler/src/compiler/RegAlloc.v
```diff
@@ -672,7 +698,8 @@ Fixpoint check(m: list (srcvar * impvar))(s: stmt)(s': stmt'){struct s}: result
     assert_in y y' m;;
     assert (Z.eqb ofs ofs') err;;
     assert (access_size_beq sz sz') err;;
-    Success (assignment m x x')
+    a <- assignment m x x';;
+    Success a
```
samuelgruetter, line 702:
```suggestion
    assignment m x x'
```
and similarly at other occurrences below

compiler/src/compiler/RegAlloc.v
```diff
@@ -1087,20 +1123,110 @@ Section CheckerCorrect.
     unfold assert_ins, assert. intros. fwd. assumption.
   Qed.
 
+  (* TODO XXX this should be obsolete *)
```
samuelgruetter, line 1126: is it obsolete indeed?

samuelgruetter (changes requested):
Looks great :tada: 

I'm adding some technical low-level comments here, and will also send some high-level comments by email.

Github said that this branch is "out of date" and "can't be merged" and also can't be rebased, and while trying to find out why, I clicked a button that directly merged master into this branch, and now everything seems to be ok. But that just means that before continuing, you'll have to do a `git pull` on your branch to get that merge commit.

> * I have only just now realized the ipow example no longer compiles, because I factored out lots of hardcoded constants into functions like `cost_op`. This should be extremely easy to fix at the tactic automation level (but I'm making this pull request now in the interest of being able to see the code sooner).

yes, thanks for making the PR early on :+1: And I'll wait with merging until you got a chance to fix this one.

> * In the previous code, the large constant bound on function calls was added twice in the semantics, once in the `exec fbody` judgment and once in the postcondition. I don't really understand why this is, but I kept them both (as `cost_call_internal` and `cost_call_external`). As far as I can tell, only one of these is needed, and putting it in either place works just as well.

Would be good to look into this one too. The name cost_call_external is misleading, because it’s used for (internal) function calls, whereas for external function calls (aka interact), cost_interact is used. If both cost_call_internal and cost_call_external are needed in the places they are currently, I’d name them cost_call_preamble and cost_call_postamble. Or, even better, if only one is needed, I’d use cost_call_internal and delete cost_call_external, because we already have cost_interact.
While looking into this, could you also see if in Pipeline.compiler_correct_wp, the initial metrics, which currently are (cost_SCall_L initial.(getMetrics)), could  just be initial.(getMetrics) instead?

pratapsingh1729:
Re the two costs for calls, if I recall correctly, I indeed added that to account for the function preamble (saving temporaries to the stack) and the function postamble (putting the temporaries back into registers). It was done in this way to simplify the `FlatToRiscv` proof, so that the metrics computation for the actual call instruction started with an initial metrics after the saving of the temporaries and then ended before restoring the temporaries---this saved some annoying arithmetic I think. We might be able to fuse them into one cost, however.

tckmn:
@samuelgruetter I think this branch should be in a pretty good state to merge now! A few notes:

* I addressed most of the comments here, with the exception of the mentions of removing `cost_SCall_L` and the `cost_call_internal` in the spilled call spec in the pipeline. These extra costs cannot be removed, because the spilling and RISC-V compilation stages add boilerplate code to the program itself. I've renamed them to `cost_spill_spec` and `cost_compile_spec` for clarity (I'm not sure if I realized this at the time, but using `cost_call_internal` for the spilling phase was extremely misleading, as I don't think it was actually semantically correct, it just happened to have numbers that worked).

* There is now only one `cost_call` (rather than both `cost_call_external` and `cost_call_internal`).

* With Andres's help, the Qed slowness in the ipow example is fixed :tada:

bedrock2/src/bedrock2/MetricCosts.v
```diff
@@ -112,10 +112,35 @@ Definition isRegZ (var : Z) : bool :=
...
+             | H : context[cost_set] |- _ => H
+             | H : context[cost_if] |- _ => H
+             | H : context[cost_loop_true] |- _ => H
+             | H : context[cost_loop_false] |- _ => H
+             end in
+    let t := type of H in
+    let t' := eval cbv [cost_interact cost_call cost_load cost_store
+      cost_inlinetable cost_stackalloc cost_lit cost_op cost_set
+      cost_if cost_loop_true cost_loop_false] in t in
+    replace t with t' in H by (symmetry; reflexivity)
```
andres-erbsen, line 138: I now wonder whether `exact (eq_refl t)` that has the implicit argument specified would be fast regardless of symmetry. That would allow us to blame unpredictable slowness on implicit-argument inference by unification, rather than conversion, which would be less dissatisfying. But perhaps we've spent enough time on this already.

samuelgruetter: yeah, I'd find it interesting to run a speed comparison between the following:
* `(symmetry; reflexivity)`
* `reflexivity`
* `(exact (eq_refl t))`
* `(exact (eq_refl t'))`

bedrock2/src/bedrock2/MetricLoops.v
```diff
@@ -0,0 +1,664 @@
...
+      { destruct (Hinv' ltac:(trivial)) as (v'&inv'&ltv'v).
+        exists (Some v'); split; trivial. (* NOTE: this [trivial] simpl-reduces [with_bottom] *)
+        exists br'; split; trivial.
+        split; intros; try contradiction.
+        eexists; split; eauto. } }
+     *)
+
+    eapply Hexit.
+  (*Qed.*)
+  Admitted.
```
samuelgruetter, line 419: Please do not admit any lemmas. If there is partial work in a lemma that you would not like to delete but also don't want to finish right now, you can use `Abort` instead of `Admitted`, so that the lemma can't be used by accident.

bedrock2/src/bedrock2/MetricLoops.v
```diff
@@ -0,0 +1,664 @@
...
+    intros t' m' l' (?&?&HH&?).
+    eexists; split; eauto.
+    split; intros; eauto.
+    specialize (HH ltac:(eauto)).
+    eapply hlist.existss_exists in HH; destruct HH as (?&?&?&?&?).
+    eexists; split; eauto.
+  Qed.
+
+=======
+>>>>>>> 94c30cc1 (made rather a mess trying to get andy's metric stuff to work; commiting so I can ask andres for help)
```
samuelgruetter, line 605: This looks like left-over merge-conflict garbage, please delete. In fact, I'm not a big fan of commented-out code anyways, so maybe it's best to delete all the commented-out code.

compiler/src/compiler/FlatToRiscvMetric.v
samuelgruetter: Please delete this file before we merge

bedrock2/src/bedrock2/MetricCosts.v
tckmn: The tactic hangs with `reflexivity` and `(exact (eq_refl t))`, but it works with `(symmetry; reflexivity)` and `(exact (eq_refl t'))`. The latter looks a little less black-magic than the former (although presumably it has the same end effect of flipping the direction of a conversion), so I changed it to `(exact (eq_refl t'))`.

bedrock2/src/bedrock2/MetricLoops.v
tckmn: Whoops, I didn't realize the WIP loop lemmas had made it into the PR; I deleted everything in that file that isn't finished (including the commented out stuff), possibly to be added back later.

tckmn: (pinging @samuelgruetter since i believe there are no remaining outstanding comments)

samuelgruetter:
Merged, thanks @tckmn :tada: 

For the record, there are still a few points that would be nice to address in a later PR:

* `cost_interact`, `cost_call`, `cost_load` etc should not take an `mc` argument. Instead, its callers should add mc to them
* "TODO XXX address inconsistency in where metrics are added"
* `locals_based_call_spec` vs `locals_based_call_spec_spilled`: are both needed?
* `cost_spill_spec` and `cost_compile_spec` should not appear in the toplevel theorems, but should be part of the costs that appear in the semantics -- or what exactly do you mean by "These extra costs cannot be removed, because the spilling and RISC-V compilation stages add boilerplate code to the program itself"?

But to avoid that this PR goes stale and to make sure others can work on top of this, I now merged this PR.

PR #430, opened 2024-08-11
Document Limited LiveVerif Automation Experienced In Crit-Bit

vfukala:
During my crit-bit tree development using Live Verification, I came across some proofs that I thought the framework should ideally be able to do on its own but actually couldn't.

I haven't figured out a way of properly fixing this, so I at least extracted the code to near-minimal examples -- those are the content of this PR.


samuelgruetter:
Thanks, good to have these documented in stand-alone files!

PR #431, opened 2024-08-25
Leakage traces

OwenConoly:
This PR will add leakage traces to the source and intermediate semantics, augment the compiler-correctness statement and proof to talk about leakage traces, and add some examples involving leakage traces.


OwenConoly:
For now, this branch only has some edits to Semantics.v.  Before I add anything else, I wanted to ask whether I be making these edits (i.e., adding leakage traces) in Semantics.v, MetricSemantics.v, or both?  Also, why are there Semantics.v and MetricSemantics.v separate, and are there plans to merge them eventually?

bedrock2/src/bedrock2/Semantics.v
```diff
@@ -93,35 +114,38 @@ Section semantics.
   Section WithMemAndLocals.
     Context (m : mem) (l : locals).
 
-    Local Notation "x <- a ; f" := (match a with Some x => f | None => None end)
-      (right associativity, at level 70).
+    Local Notation "' x <- a ; f" := (match a with Some x => f | None => None end)
+      (right associativity, at level 70, x pattern).
 
-    Fixpoint eval_expr (e : expr) : option word :=
+    Fixpoint eval_expr (e : expr) (k : ltrace) : option (word * ltrace) :=
```
OwenConoly, line 120: Is this function nice as is (taking leakage trace as argument), or should I rewrite it to have type `expr -> option (word * ltrace)`?

bedrock2/src/bedrock2/Semantics.v
```diff
@@ -14,12 +14,22 @@ Definition LogItem{width: Z}{BW: Bitwidth width}{word: word.word width}{mem: map
...
   list LogItem.
 
+Inductive event {width: Z}{BW: Bitwidth width}{word: word.word width} : Type :=
+| leak_unit
+| leak_bool (b : bool)
+| leak_word (w : word)
+| leak_list (l : list word).
+(* sometimes it's convenient that one io call leaks only one event
+   See Interact case of spilling transform_trace function for an example. *)
+Definition ltrace {width: Z}{BW: Bitwidth width}{word: word.word width} : Type :=
```
OwenConoly, line 24: Is the name `ltrace` good as is, or should I write out `leakage_trace`?  Or name it something else entirely?

andres-erbsen (commented):
> Semantics.v, MetricSemantics.v, or both?

IMO the best options is to create *additional* semantics files. We want to have only one compiler proof, so that would use LeakageMetricSemantics. Some program proofs may be done with that one as well, but the examples for our current manuscript probably shouldn't be, so there we'd probably want want LeakageSemantics as well. In an unverified software project, this kind of duplication would be asking for inconsistency bugs, but I hope it won't be too hard to state and prove the appropriate relationships between these variants.

OwenConoly:
> I hope it won't be too hard to state and prove the appropriate relationships between these variants.

Proving the appropriate relationship between Semantics.v and LeakageSemantics.v will only be trivial if we modify Semantics.v so that `exec` takes the `pick_sp` parameter.  If LeakageSemantics `exec` takes `pick_sp` as a parameter but Semantics `exec` stays as is, then proving any sort of equivalence between them is equally as nontrivial as that semantics-equivalence statement we put in the paper, the one that I wrote a ~3000-line proof of.

So, should I modify Semantics and MetricSemantics `exec`s to take `pick_sp` as a parameter, and then prove the appropriate relationships between Semantics vs. MetricLeakageSemantics, MetricSemantics vs. MetricLeakageSemantics, and LeakageSemantics vs MetricLeakageSemantics?

OwenConoly:
However, the Semantics-`exec`-implies-LeakageSemantics-`exec` direction is trivial even if LeakageSemantics takes `pick_sp` as a parameter and Semantics doesn't.  So if that direction of the equivalence is all we care about, then potentially I could just leave Semantics and MetricSemantics alone.

andres-erbsen:
Ah, indeed. I was only thinking of the easy direction, Semantics implies LeakageSemantics. I don't think we should add pick_sp to current semantics. @samuelgruetter do you think it would be acceptable to carry both copies in the repo, but only the half of the equivalence proof that allows program proofs without leakage to be used with a compiler proof that preserves everything they say about leakage?

However, the other direction of this relationship might be a good opportunity to drill down on the difficulty in the equivalence proof and perhaps simplify it. We don't *have* to do this, but I'm thinking that if you could state the desired equivalence with a minimized version of semantics that is just complicated enough to illustrate the challenge (without abstract traces, predictors, or even  function calls and etc), then I would give it a try  "How hard can pushing a quantifier into an inductive be" :rocket::joy::roll_eyes:

OwenConoly:
> "How hard can pushing a quantifier into an inductive be"

This could be nice, yeah!  To make it as academic as possible (and remove the coqutil/bedrock2 dependency and get a standalone file) I'd probably make the following simplifications to the equivalence proof I currently have:
- no memory, only locals
- no leakage traces or metrics either
- locals are nats instead of words (in addition to removing a dependency, this makes me happy because proofs seem less trivial when they're written about Turing-complete languages)
- no ifs, since you can write them with whiles anyway
- instead of stack allocation, the quantifier is just an RNG construct
- no function calls

I'd probably keep IO calls though, since the nondeterminism there seems to add an interesting dimension.

OwenConoly:
Oh, bedrock2 is Turing-complete!  I forgot that you can store and retrieve as much information as you want in the IO trace (with an appropriate choice of `ext_spec`).  That's kinda cool.

andres-erbsen:
:+1:  You can store a turing-machine tape in a nat, so I think all options here are turing-complete. And I'd even remove I/O at first, we can add it back later if we want.

OwenConoly:
The existence of (Metric)LeakageSemantics.v implies the existence of:

(Metric)LeakageLoops.v, (Metric)LeakageProgramLogic.v, (Metric)LeakageWeakestPreconditionProperties.v, and (Metric)LeakageWeakestPrecondition.v,

right?

andres-erbsen:
Yes. At some point I may try to share more code, but for new these are just copied I think.

samuelgruetter:
> The existence of (Metric)LeakageSemantics.v implies the existence of:
> 
> (Metric)LeakageLoops.v, (Metric)LeakageProgramLogic.v, (Metric)LeakageWeakestPreconditionProperties.v, and (Metric)LeakageWeakestPrecondition.v,
> 
> right?

Will we ever need a program logic that does both metrics and leakage at the same time? If not, the needed files should probably just be

* MetricLeakageSemantics.v (because in the compiler, we want both at the same time)
* (non-metric) LeakageLoops.v, LeakageProgramLogic.v, LeakageWeakestPreconditionProperties.v, and LeakageWeakestPrecondition.v

OwenConoly:
Ok.  I can just do the MetricLeakage files while I'm at it, if they'll be needed eventually anyway.

Unrelatedly, since I am adding the separate files for Leakage stuff instead of changing what was already there, my changes to the source language shouldn't break the compiler proof.  So it seems natural to split this PR into two: in this PR I can just change the source language and add some examples, and then I can make a separate PR for compiler proof things.  Does that sound good?

andres-erbsen:
Separate PRs sounds good to me. (There is some possibility that we will want to review them together, though.)

compiler/src/compiler/CustomFix.v
```diff
@@ -0,0 +1,134 @@
...
+Check Fix_eq'.
+Definition type_of_Fix_eq' :=
+  forall (A : Type) (R : A -> A -> Prop) (Rwf : well_founded R) 
+    (P : A -> Type) (F : forall x : A, (forall y : A, R y x -> P y) -> P x),
+    (forall (x : A) (f g : forall y : A, R y x -> P y),
+        (forall (y : A) (p1 p2 : R y x), f y p1 = g y p2) -> F x f = F x g) ->
+    forall x : A, Fix Rwf P F x = F x (fun (y : A) (_ : R y x) => Fix Rwf P F y).
+                               
+Goal type_of_Fix_eq' -> type_of_Fix_eq.
+  cbv [type_of_Fix_eq type_of_Fix_eq']. auto. Qed.
```
OwenConoly, line 46: `Fix_eq` was apparently useless for proving the fixpoint equation for some functions that I defined for the compiler proof, so I proved and used this `Fix_eq'` instead.  I'm unsure whether `Fix_eq` is genuinely not applicable here; it's probable that I'm just not using `Fix`/`Fix_eq` as intended.  If `Fix_eq` is really not helpful here, then I wonder why they don't have `Fix_eq'` in the standard library.

I think that, a few months ago, I had a good grasp on why `Fix_eq` is not helpful for the function I defined and why `Fix_eq'` is needed instead.  But my current level of understanding is just "if I apply `Fix_eq` instead of `Fix_eq'`, then the proof gets stuck".

@andres-erbsen, you've directly used `Fix` for general recursion, right?  Has this issue come up before?

andres-erbsen: I have run into this before, and I think I did get it working with stdlib Fix_eq every time. Unfortunately I do not recall how, but perhaps the trick was about clearing the recursive call before the abstract proof that R decreases. However, I remember spending at least a few minutes on it and feeling annoyed that I had to, so feel free to keep the more flexible lemma and perhaps even submit it to stdlib.

OwenConoly:
Unfortunately, my attempt at writing my compiler-proof leakage-transformation functions without CPS did not work out nicely, for two reasons:
- It's convenient to be able to do some error-handling-ish logic, where hitting a certain thing in the recursion causes the function to exit immediately.  This is kind of annoying to do without CPS: the solution I found was to carry around a flag saying 'have we hit an error yet'.  This isn't a big problem though, and on its own probably wouldn't make using CPS worth it.
- The bigger issue is that, because we have the deterministic stack allocation addresses, our inductive hypothesis for the compiler phase proofs needs to say that the high-level pick_sp function is of a particular form.  With the function in CPS, there's a natural way to say what this form is, and it's easy to prove given the recursive structure of the continuation-passing function.  With the function not in CPS, it is not so convenient to say what it should be.  Basically, the flexibility lost is that in CPS, we can just state that the high-level pick_sp function should look like the leakage transformation function, but it doesn't matter what the continuation is.

Amusingly, this second issue is a non-issue if we work with nondeterministic stackalloc addresses.  This made me assume it wouldn't be an issue with the deterministic stackalloc addresses, but apparently I was wrong.

I probably didn't explain any of that well enough to be understandable; I could elaborate if any of it is interesting.  But anyway, since I have to use CPS now, I have a need to write a non-structurally-recursive function which takes a function as an argument.  I think  there are three options for how to prove the fixpoint equation for such a function:
1. Use a functional extensionality axiom.
2. Prove some analogue of Fix_eq that's parameterized over arbitrary equivalence relations instead of just equality.
3. Do what is being done in [fiat/src/Common/Wf1.v](https://github.com/mit-plv/fiat/blob/4063f04f5c3605ad06b8a476e2300505cab1ca5d/src/Common/Wf1.v#L313).  I don't know exactly what is going on there, but it looks like it solves this problem?

I already did option (2), so I'm inclined to just stick with that unless there are compelling reasons for doing something else.  I imagine there are no good reasons for doing (1), but I'm not sure about (3).

compiler/src/compiler/CustomFix.v
```diff
@@ -1,165 +1,144 @@
...
+  intro x; unfold Fix, Wf.Fix.
+  rewrite <- Fix_F_eq'.
+  intros. apply F_ext; intros.
+  - assumption.
+  - apply Fix_F_inv'. assumption.
+Qed.
+End GeneralFixPoint.
 
 (*did jason do this already?  should explore
   https://github.com/mit-plv/fiat/blob/master/src/Common/Wf1.v*)
```
JasonGross, line 49: If you're looking for Fix_eq that supports functions of many (non-recursive) arguments (following the recursive argument), or Fix that recurses simultaneously on many variables (via tuples), then indeed the theory is already present in Wf1 / Wf2.  But it looks like maybe you're doing something different?

JasonGross: The relevant place to look to understand Wf1, btw, is starting around https://github.com/mit-plv/fiat/blob/4063f04f5c3605ad06b8a476e2300505cab1ca5d/src/Common/Wf1.v#L408, where we instantiate the general lemmas to construct all the variants

JasonGross:
> 3\. Do what is being done in [fiat/src/Common/Wf1.v](https://github.com/mit-plv/fiat/blob/4063f04f5c3605ad06b8a476e2300505cab1ca5d/src/Common/Wf1.v#L313).  I don't know exactly what is going on there, but it looks like it solves this problem?
> 
> 
> I already did option (2), so I'm inclined to just stick with that unless there are compelling reasons for doing something else. I imagine there are no good reasons for doing (1), but I'm not sure about (3).

I'm not sure.  Wf1 is a complicated general wrapper around the simple trick that if you know how many arguments your recursive function takes, you can demand pointwise extensionality rather than equality, and the induction goes through.  It sounds like what you did is somewhat more general than this (though note that you maybe want to parameterize over a PER instead of an equivalence, if you want to avoid funext?)

andres-erbsen:
I don't have opinions on this, feel free to do whatever.

andres-erbsen:
As discussed in person, this branch is in overall good shape and seems desirable to merge it soon rather than to wait for perfection.

@samuelgruetter do you have any concerns?

compiler/src/compiler/Pipeline.v
```diff
@@ -155,50 +184,109 @@ Section WithWordAndMem.
...
       exists argnames retnames fbody l,
         map.get e f = Some (argnames, retnames, fbody) /\
         map.of_list_zip argnames argvals = Some l /\
-        Exec e fbody t m l mc (fun t' m' l' mc' =>
+        Exec pick_sp e fbody k t m l mc (fun k' t' m' l' mc' =>
                        exists retvals, map.getmany_of_list l' retnames = Some retvals /\
-                                       post t' m' retvals mc').
+                                       post k' t' m' retvals mc').
+
+    Lemma thing_locals_based_call_spec_weaken{Var Cmd: Type}{locals: map.map Var word}
```
andres-erbsen, line 217: Please consider a rename :)

compiler/src/compiler/Pipeline.v
```diff
@@ -64,10 +68,18 @@ Section WithWordAndMem.
...
-        (t: trace)(m: mem)(argvals: list word)(mc: MetricLog)
-        (post: trace -> mem -> list word -> MetricLog -> Prop): Prop;
+      Valid: Program -> Prop;
+      Settings: Type;
+      SettingsInhabited: Settings;
+      Leakage: Type;
+      Call(pick_sp: list Leakage -> word)(p: Program)(s: Settings)(funcname: string)
+        (k: list Leakage)(t: trace)(m: mem)(argvals: list word)(mc: MetricLog)
+        (post: list Leakage -> trace -> mem -> list word -> MetricLog -> Prop): Prop;
+      WeakenCall: forall pick_sp p s funcname k t m mc argvals (post1: _ -> _ -> _ -> _ -> _ -> Prop),
```
andres-erbsen, line 78: @samuelgruetter do properties of Call belong here?

compiler/src/compilerExamples/memequal.v
```diff
@@ -0,0 +1,555 @@
...
+      eapply WeakestPreconditionProperties.Proper_call.
+      2: eapply spec.
+      cbv [Morphisms.pointwise_relation Basics.impl]. intros.
+      destruct H8 as [ [k'' [H8 H9] ] [H10 [H11 H12] ] ]. subst.
+      split; [|reflexivity].
+      destruct H12 as [k''_ [H12_1 H12_2] ]. subst.
+      rewrite app_nil_r. reflexivity. }
+  Qed.*)
+
+  Check (@compiler_correct_wp _ _ Words32Naive.word mem _ ext_spec _ _ _ ext_spec_ok _ _ _ _ _ word_ok _ _ RV32I _ compile_ext_call leak_ext_call compile_ext_call_correct compile_ext_call_length fs_swap instrs_swap finfo_swap req_stack_size_swap fname_swap _ _ _ _).
```
andres-erbsen, line 341: Either remove this or turn it into a lemma.

etc/test_itauto.sh
```diff
@@ -23,10 +23,6 @@ make
 make install
 cd ../..
 
-cd deps/coq-record-update/
```
andres-erbsen, line 26: What's going on here?

samuelgruetter (commented):
LGTM (modulo Andres' and my review comments)

compiler/src/compiler/Pipeline.v
```diff
@@ -64,10 +68,18 @@ Section WithWordAndMem.
 
   Record Lang := {
     Program: Type;
-    Valid: Program -> Prop;
-    Call(p: Program)(funcname: string)
-        (t: trace)(m: mem)(argvals: list word)(mc: MetricLog)
-        (post: trace -> mem -> list word -> MetricLog -> Prop): Prop;
+      Valid: Program -> Prop;
+      Settings: Type;
+      SettingsInhabited: Settings;
```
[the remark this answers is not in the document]
OwenConoly: Right, yes; I wrote this the way I did just because it seemed weird to say that two functions with, say, different return addresses are in different "languages".  Certainly the partial application would be less verbose.  I'll do that instead.

compiler/src/compiler/FlatToRiscvFunctions.v
```diff
@@ -385,7 +392,45 @@ Section Proofs.
   Hint Extern 3 (map.only_differ _ _ _)
   => eapply only_differ_trans_l; [eassumption|eauto with map_hints ..]
   : map_hints. *)
+  Search option_map.
```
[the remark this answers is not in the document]
OwenConoly: Actually I've discovered that neither of these lemmas was being very useful anywhere---`option_map_option_map` wasn't doing anything, and I was able to replace `option_map_option_map'` by something less hacky---, so I've just deleted them.

bedrock2/src/bedrock2/BasicC32Semantics.v
```diff
@@ -1,5 +1,5 @@
 Require Import Coq.ZArith.ZArith.
-Require Import bedrock2.Syntax bedrock2.Semantics.
+Require Import bedrock2.Syntax bedrock2.Semantics bedrock2.LeakageSemantics.
```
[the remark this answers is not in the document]
OwenConoly: I think I did this because `ct.v` uses `BasicC32Semantics`.  Should I make it use `FE310CSemantics` instead?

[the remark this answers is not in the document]
OwenConoly: Update: I just changed it to use FE310CSemantics instead.

andres-erbsen:
yes

bedrock2/src/bedrock2/LeakageWeakestPrecondition.v
```diff
@@ -0,0 +1,326 @@
...
+              (exists r0,
+                  .. (exists rn,
+                         rets = (cons r0 .. (cons rn nil) ..) /\
+                         post) ..))))
+    (at level 200,
+     name at level 0,
+     r0 closed binder, rn closed binder,
+     tr name, tr' name, mem name, mem' name,
+     pre at level 200,
+     post at level 200).
```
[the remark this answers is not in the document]
OwenConoly: Yes, I believe they did not work.  But now I think I have fixed them.

There are now two notations in LeakageWeakestPrecondition.  There is the `fnspec!` notation, which is just like the `fnspec!` in WeakestPrecondition.

There is also the `fnspec_ex!` notation, which is just like `fnspec!` except that an existential quantifier has been inserted in the appropriate place for stating that a program is constant-time.  Do you like this notation?  There are now many examples of it being used in bedrock2Examples/ct.v.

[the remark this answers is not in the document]
andres-erbsen: Hmm, they sure seem better than no notation. How about `fnspec! exists f ... g, "getline" x y z ~> v, { ...`? I believe "exists" is already a keyword, and perhaps this way a user familiar with fnspec! will guess the meaning of the new notation correctly more often.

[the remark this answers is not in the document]
OwenConoly: That seems nice and intuitive.  I changed the notations to look like that instead.

OwenConoly:
This branch is still some distance (not sure how far) from being in a state where it compiles.  But I'll mention as a future TODO, probably not for this PR, that it would be nice to have a constant-time-related compiler example that does something with IO.

OwenConoly:
Currently, e1000_read_write_step.v does not compile, because it uses LeakageSemantics `exec`, but it is pulling its `ext_spec` from memory_mapped_ext_spec.v, which is a LeakageSemantics-style `ext_spec`.

(I changed memory_mapped_ext_spec.v to have a LeakageSemantics `ext_spec` because that is what the compiler proof wants.)

Should I fix this by changing e1000_read_write_step.v to use LeakageSemantics `exec`?

Or should e1000_read_write_step.v continue using Semantics `exec`?  Then it would have to use some Semantics-style `ext_spec` which is related to that of memory_mapped_ext_spec.v in the appropriate way.

andres-erbsen:
I think porting the example to use LeakageSemantics throughout shouldn't be an issue; and it seems easier than creating an appropriate abstraction to insulate it from the change. wdyt @samuelgruetter ?

samuelgruetter:
> I think porting the example to use LeakageSemantics throughout shouldn't be an issue; and it seems easier than creating an appropriate abstraction to insulate it from the change. wdyt @samuelgruetter ?

I was expecting/hoping that similarly to MetricSemantics, it should still be possible to write developments that do not care about leakage. Is that not the case or deemed to hard?

samuelgruetter:
Could you define a generic wrapper that takes a `Semantics.ExtSpec` and turns it into a `LeakageSemantics.ExtSpec` that makes no guarantees about leakage, but has the right type?

OwenConoly:
Sure!  It does seem nicer to go the other direction, though---having a wrapper `LeakageSemantics.ExtSpec -> Semantics.ExtSpec`---since then the `LeakageSemantics.ExtSpec` instance can potentially be stronger (i.e., put a stronger restriction on the external thing that the bedrock2 program is interacting with).

So, should I define both a `LeakageSemantics.ExtSpec` and a `Semantics.ExtSpec` in memory_mapped_ext_spec.v, one in terms of the other via a wrapper?  Do you have a preference for the direction of the wrapper?

bedrock2/src/bedrock2/MetricLeakageSemantics.v
```diff
@@ -0,0 +1,709 @@
...
+      destruct IHarges as (mc'' & IH). rewrite IH. eauto.
+  Qed.
+
+  Context (e: env).
+  Context (sem_ext_spec : Semantics.ExtSpec := fun t mGive a args post => ext_spec t mGive a args (fun mReceive resvals klist => post mReceive resvals)).
+  Existing Instance sem_ext_spec.
+
+  Lemma to_plain_exec: forall c t m l post,
+      Semantics.exec e c t m l post ->
+      forall k mc, exec e c k t m l mc (fun _ t' m' l' _ => post t' m' l').
```
OwenConoly, line 649: Here's what the `LeakageSemantics.ExtSpec -> Semantics.ExtSpec` wrapper would look like.

samuelgruetter: I am confused about what you mean by "direction": This lemma seems to be a mix of both directions: For the ext_spec, it is leakage->plain, but for the exec, it is plain->leakage? Can you make a wrapper that is in plain->leakage direction for both?

OwenConoly: Yeah, I'm referring to the `ext_spec` direction, i.e., the type of the ext-spec wrapper in the code snippet is `LeakageSemantics.ExtSpec -> Semantics.ExtSpec`.

I think it would be easy to make (and prove correct) a wrapper of type `Semantics.ExtSpec -> LeakageSemantics.ExtSpec`.  Do you think that would be preferable to having a wrapper of type `LeakageSemantics.ExtSpec -> Semantics.ExtSpec`?

samuelgruetter: Yes, a wrapper of type `Semantics.ExtSpec -> LeakageSemantics.ExtSpec` is preferable, because it should enable you to reprove the original plain compiler-correctness theorem using the new compiler-correctness theorem with leakage, and then you don't need to change code that was using the original compiler-correctness theorem.

OwenConoly: Oh, I see!  Yes, that's an option.  I had assumed y'all didn't want to deal with multiple versions of the compiler-correctness theorem, since there were no separate MetricSemantics and Semantics compiler-correctness theorems.

I ended up implementing the `LeakageSemantics.ExtSpec -> Semantics.ExtSpec` wrapper and corresponding proof, since I needed it anyway for FE310CSemantics.  Then it was trivial to add a `Semantics.ExtSpec` instance to memory_mapped_ext_spec, so the e1000* files can stay as they are.

So, the current status is that memory_mapped_ext_spec has two `ext_spec`s: one with leakage, and one without.  The leakage one is used with the compiler proof (in memory_mapped_ext_calls_compiler).  The non-leakage one is used in all the source-level files with prefix e1000.  The two `ext_spec`s are related via the `to_plain_exec` lemma in the snippet above.  Does this seem like a good state of affairs?

OwenConoly:
This branch compiles on my machine now.  I'll take a pass through it soon to fix typos, unneeded comments, Print statements, etc., and after that it should be about ready to merge.

bedrock2/src/bedrock2Examples/stackalloc.v
```diff
@@ -18,24 +18,66 @@ Definition stackdisj := func! () ~> (a,b) {
...
+      rewrite Znat.Z2Nat.id.
+      + replace ((width + 7) / 8 * 8) with (width + 7 - (width + 7) mod 8).
+        -- assert (H := Z.mod_pos_bound (width + 7) 8). blia.
+        -- rewrite Zdiv.Zmod_eq_full; blia.
+      + apply Z.div_pos; try blia. destruct word_ok. blia.
+  Qed.
+
+  Lemma word_to_bytes' (a : word) :
+    exists l, length l = (Z.to_nat ((width + 7) / 8)) /\
+                a = word.of_Z (LittleEndianList.le_combine l).
```
OwenConoly, line 65: Thoughts about where this belongs?  [coqutil/Word/Properties.v](https://github.com/mit-plv/coqutil/blob/master/src/coqutil/Word/Properties.v), perhaps?

OwenConoly: Actually, [coqutil/Word/LittleEndianList.v](https://github.com/mit-plv/coqutil/blob/master/src/coqutil/Word/LittleEndianList.v) seems better.

andres-erbsen: Are you sure you want a lemma with an `exists` litke this at all? Probably a lemma that uses a specific value of `l` would do, and maybe it exists already.

OwenConoly: Yeah, `word_to_bytes'` is a corollary of `word_to_bytes` (right above it), which has a specific value of `l`.  I just thought it might be nice to abstract away the unnecessary details, since (so far as I know) this lemma would mainly be useful for stack deallocation, where we wouldn't care what the bytes actually are.

I don't think a lemma like this existed before I proved it.  I just conducted an unsuccessful three-minute search, and also [this comment](https://github.com/mit-plv/bedrock2/blob/b2cd2baf28b89c0b29e2f98bcd7721b90adbdd66/bedrock2/src/bedrock2Examples/indirect_add.v#L264) makes it sound like there are no word_to_bytes lemmas.

compiler/src/compiler/Pipeline.v
```diff
@@ -252,6 +257,18 @@ Section WithWordAndMem.
...
+    Definition remove_n_r {X : Type} (n : nat) (l : list X) :=
+      rev (skipn n (rev l)).
+
+    Lemma remove_n_r_spec {X : Type} (l1 l2 : list X) :
+      remove_n_r (length l2) (l1 ++ l2) = l1.
+    Proof.
+      cbv [remove_n_r]. rewrite rev_app_distr. rewrite List.skipn_app_r.
+      - apply rev_involutive.
+      - rewrite length_rev. reflexivity.
+    Qed.
```
OwenConoly, line 270: This probably belongs somewhere other than Pipeline.v.  Shall I put it in coqutil?

OwenConoly:
compiling with 8.18 now works on my machine

PR #435, opened 2024-11-13
[do not merge] Metriclightbulb

samuelgruetter:
Opening a PR to get CI's judgment on what compiles and what doesn't -- locally (with Coq 8.18.0 after replacing `length_app` by `app_length`) I got an error in `CompilerInvariant.v, line 180, characters 36-38` because `compile_prog` now seems to take two additional string arguments `init_f` and `loop_f`


andres-erbsen:
I haven't compiled anything other than MetricLightbulbInvariant, on 8.20. For eventually merging this, we probably want to move that file to compiler examples and duplicate ToplevelLoop.v into old and new version.

PR #441, opened 2025-01-04
Move separation-logic basics into coqutil.

jadephilipoom:
Follow-up to mit-plv/coqutil#124
- First commit bumps coqutil to catch the copied files
- Second commit removes the files from bedrock2 and replaces references to them with references to coqutil.

To minimize the changes I kept the files in bedrock2 but made them simply one-line exports of the new locations in coqutil.


jadephilipoom:
CC @andres-erbsen

andres-erbsen:
> build (master) Expected — Waiting for status to be reported

PR #444, opened 2025-01-30
Add verified shrd, aka funnel shift right.

lukaszobernig:
We prove two specs: One that uses bitwise operations, and one that uses arithmetic operations.


andres-erbsen:
image

PR #445, opened 2025-01-31
Add some lemmas relating the different semantics

OwenConoly:
In addition to what the title says, this PR makes it so that `LeakageProgramLogic.straightline_call` works even when the available function spec is written in terms of `Semantics.exec` rather than `LeakageSemantics.exec`.

This is needed to fix mit-plv/fiat-crypto#2007.


bedrock2/src/bedrock2/MetricLeakageSemantics.v
```diff
@@ -546,173 +546,4 @@ Section WithParams.
     - eassumption.
     - eapply exec.extend_env; eassumption.
   Qed.
-
```
andres-erbsen, line 549: Does this mean that a `Require` could be removed?

OwenConoly: Not sure which one you were talking about, but I removed a few that were obviously not needed.  (my pushing that disabled your auto-merge thing)

PR #457, opened 2025-05-08
[DO NOT MERGE] Smarter ToCString

andres-erbsen:
- less parentheses around variables, literals, unary operators
- typedef for `br_word_t`
- per-size load/store functions/macros
- compiler-specific no-strict-aliasing tricks in prelude

PR #459, opened 2025-05-28
less ambitious ToCString improvements

andres-erbsen:
Less ambitious alternative to https://github.com/mit-plv/bedrock2/pull/457

Cc @davidben

### DIFF
--- a/bedrock2/src/bedrock2/ToCString.v
+++ b/bedrock2/src/bedrock2/ToCString.v
@@ -1,71 +1,105 @@
 Require Import bedrock2.Syntax bedrock2.Variables. Import bopname.
 Require Import coqutil.Datatypes.ListSet.
-Require Import Coq.ZArith.BinIntDef Coq.Numbers.BinNums Coq.Numbers.DecimalString.
-Require Import Coq.Strings.String. Local Open Scope string_scope.
+Require Import Coq.ZArith.BinIntDef Coq.Numbers.BinNums Coq.Numbers.DecimalString Strings.HexString.
+Require Import Coq.Strings.String. Import Byte. Local Open Scope string_scope.
 
-Definition prelude := "#include <stdint.h>
+Definition prelude := "// Generated from Bedrock code. Avoid editing directly.
+#include <stdint.h>
 #include <string.h>
 #include <assert.h>
 
-static __attribute__((constructor)) void _br2_preconditions(void) {
-  static_assert(~(intptr_t)0 == -(intptr_t)1, ""two's complement"");
-  assert(((void)""two's complement"", ~(intptr_t)0 == -(intptr_t)1));
-  uintptr_t u = 1;
-  assert(((void)""little-endian"", 1 == *(unsigned char *)&u));
-  intptr_t i = 1;
-  assert(((void)""little-endian"", 1 == *(unsigned char *)&i));
+#define BR_WORD_MAX UINTPTR_MAX
+typedef uintptr_t br_word_t;
+typedef intptr_t br_signed_t;
+
+static_assert(sizeof(br_word_t) == sizeof(br_signed_t), ""signed size"");
+static_assert(UINTPTR_MAX <= BR_WORD_MAX, ""pointer fits in int"");
+static_assert(~(br_signed_t)0 == -(br_signed_t)1, ""two's complement"");
+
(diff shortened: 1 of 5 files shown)


bedrock2/src/bedrock2/ToCString.v
```diff
@@ -1,126 +1,176 @@
...
-  | expr.ite c e1 e2 => "(" ++ c_expr c ++ ") ? (" ++ c_expr e1 ++ ") : (" ++ c_expr e2 ++ ")"
+  | expr.load s ea => "_br_load" ++ c_size s++ "(" ++ c_expr ea ++ ")"
+  | expr.inlinetable s t index => "({ static const uint8_t _inlinetable["++c_lit (Z.of_nat (List.length t))++"] = {"++concat ", " (List.map HexString.of_N (List.map Byte.to_N t))++"}; _br_load"++c_size s++"((br_word_t)&_inlinetable["++c_expr index++"]); })"
+  | expr.op1 op e1 => c_op1 op (p_expr e1)
+  | expr.op op e1 e2 =>
+    match op with
+    | slu | sru | srs => c_bop (p_expr e1) op (p_shamt e2)
+    | add | sub | mul
+    | and | or  | xor => c_bop (p_expr e1) op (p_conv e2)
+    |_  => c_bop (p_expr e1) op (p_expr e2)
```
davidben, line 169: Guessing this is the style?
```suggestion
    | _  => c_bop (p_expr e1) op (p_expr e2)
```

bedrock2/src/bedrock2/ToCString.v
```diff
@@ -1,126 +1,176 @@
...
-  | access_size.word => "sizeof(uintptr_t)"
+  | access_size.word => ""
   end.
 
 Fixpoint c_expr (e : expr) : string :=
+  let p_expr e := match e with (* parenthesized expression *)
+                  | expr.literal _ | expr.var _ => c_expr e
+                  | expr.op1 _ _ (* MSVC *) | _ => "(" ++ c_expr e ++ ")"
+                  end in
+  let p_conv e := match e with (* usual arithmetic conversions to br_word_t *)
```
davidben, line 148: Took me a bit to understand what "usual arithmetic conversions to br_word_t" meant. Perhaps (if I'm understanding this right) something like (feel free to wordsmith):

> `e` is passed to a binary operand whose other parameter is a br_word_t, so if it is a literal, we can avoid a cast. This lets us emit code as `a+5` instead of `a+(br_word_t)5`.

bedrock2/src/bedrock2/ToCString.v
```diff
@@ -147,12 +197,11 @@ Definition c_act := c_call.
...
-    => indent ++ "_br2_store(" ++ c_expr ea ++ ", " ++ c_expr ev ++ ", " ++ c_size s ++ ");" ++ LF
+    => indent ++ "_br_store" ++ c_size s ++ "(" ++ c_expr ea ++ ", " ++ c_expr ev ++ ");" ++ LF
   | cmd.stackalloc x n body =>
-    let tmp := "_br2_stackalloc_"++x in (* might shadow if not fresh, likely type error... *)
-    indent ++ "{ uint8_t "++tmp++"["++c_lit n++"] = {0}; "++x++" = (uintptr_t)&"++tmp++";"++LF++
-    c_cmd indent body ++
-    indent ++ "}" ++ LF
+    let tmp := "_br_stackalloc_"++x in (* might shadow if not fresh, likely type error... *)
+    indent ++ "uint8_t "++tmp++"["++c_lit n++"] = {0}; "++x++" = (br_word_t)&"++tmp++";"++LF++
+    c_cmd indent body
```
davidben, line 204: This probably warrants a comment to explain the quirk. Perhaps:

> To be more correct, the output should be in its own block, but this makes output harder to read. Extending the lifetime of the stack allocation still produces a correct program. This assumes that stackalloc names are unique within a function, otherwise we may output something the C compiler rejects.
>
> If this becomes a problem, we can include the braces when `c` is the left half of a `cmd.seq`.

(Feel free to wordsmith. I've certainly noticed the prevailing style here is to... have far fewer comments than my suggestion. :-) )

davidben (commented):
Generally looks pretty reasonable!

PR #482, opened 2025-08-26
Add CI test against latest Debian testing (currently 8.20)

andres-erbsen:


andres-erbsen:
To be merged if all CI jobs pass including the new one.

PR #498, opened 2025-10-29
Adapt to rocq-prover/rocq#21241 (eq_rec not defined using eq_rect)

SkySkimmer:


ppedrot:
Ping, this should be backwards compatible and mergeable as-is.

PR #509, opened 2026-01-27
Adapt to https://github.com/rocq-prover/rocq/pull/21609

proux01:
Adapt to https://github.com/rocq-prover/rocq/pull/21609


proux01:
Yes, this is now an overlay for https://github.com/rocq-prover/rocq/pull/21609 and as you can see the discussion there may take some time. So you can either wait and see whether you need it or just merge it and forget about it alltogether.

PR #533, opened 2026-05-22
Adapt to rocq-prover/rocq#21987 (secvar status)

SkySkimmer:


compiler/src/compiler/ForeverSafe.v
```diff
@@ -98,7 +101,7 @@ Section ForeverSafe.
     - eapply mcomp_sat_weaken. 2: eassumption.
       intros.
       eapply H0. assumption.
-  Fail Qed. Abort. (* TODO report *)
```
SkySkimmer, line 101: I guess this TODO report was about some section variable bug fixed by rocq-prover/rocq#21987

bedrock2/src/bedrock2/LeakageWeakestPreconditionProperties.v
```diff
@@ -21,9 +21,9 @@ Section WeakestPrecondition.
        when trying to use Proper_load, or, due to COQBUG https://github.com/coq/coq/issues/11487,
        we'd get a typechecking failure at Qed time. *)
     repeat match goal with x : ?T |- _ => first
-      [  constr_eq T X; move x before pick_sp
-       | constr_eq T X; move x before ext_spec
-       | constr_eq T X; move x before locals
+      [  constr_eq x pick_sp
+       | constr_eq x ext_spec
+       | constr_eq x locals
```
SkySkimmer, line 26: this is basically emulating a "is_section_variable" tactic, I could have added one in rocq-prover/rocq#21987 but we would need this code for backwards compat anyway so I think it will wait for another PR

PR #539, opened 2026-07-14
Implement and verify u320_add function to add two 320-bit numbers.

klinashka:
This PR verifies a 320-bit addition function.  

The specification assumes the input and output words do not alias as a precondition and asserts that the output word contains the least significant 320 bits of the addition and that it returns the output bit.


andres-erbsen (approved):
Looks good! I'm not sure whether it makes more sense to develop the rest of the modinv function in fiat-crypto or in bedrock2, but I trust Miriam's judgement on that.

PR #551, opened 2026-08-17
Refactor BasicCSemantics

miriampolzer:
This moves BasicCSemantics from a global typeclass based approach to a hint based approach.
You now need to explictly Import the respective semantics file to get the right typeclasses, a simple Require does not suffice anymore.
The main advantage is that each file now needs to be very explicit about which word size to use, and the hint-based instances instead of additional definitions allow for good interaction between word-size agnostic files and files with specific word sizes. I plan to use this in the P256 formalization.

This depends on https://github.com/mit-plv/coqutil/pull/179. It needs to be merged with the bump of coqutil after merging.

### DIFF
--- a/bedrock2/src/bedrock2/BasicCSemantics.v
+++ b/bedrock2/src/bedrock2/BasicCSemantics.v
@@ -1,0 +1,85 @@
+Require Import Coq.ZArith.ZArith.
+Require Import bedrock2.Syntax bedrock2.Semantics.
+Require coqutil.Datatypes.String coqutil.Map.SortedList coqutil.Map.SortedListString.
+Require Import coqutil.Word.Interface coqutil.Map.SortedListWord.
+Require Import coqutil.Word.Naive.
+
+(* Local because it automatically adds arbitrary word size instance. *)
+#[local] Hint Extern 0 (word.word ?width) => exact (Naive.word width) : typeclass_instances.
+
+#[export] Instance word_ok {width} {BW : Bitwidth.Bitwidth width}: word.ok (Naive.word width).
+Proof. destruct Bitwidth.width_cases as [W|W]; symmetry in W; destruct W; [exact word32_ok | exact word64_ok]. Defined.
+
+#[export] Hint Extern 0 (Interface.map.map (@word.rep _ (Naive.word _)) Coq.Init.Byte.byte) =>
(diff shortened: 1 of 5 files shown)


bedrock2/src/bedrock2/BasicCSemantics.v
```diff
@@ -0,0 +1,85 @@
+Require Import Coq.ZArith.ZArith.
+Require Import bedrock2.Syntax bedrock2.Semantics.
+Require coqutil.Datatypes.String coqutil.Map.SortedList coqutil.Map.SortedListString.
+Require Import coqutil.Word.Interface coqutil.Map.SortedListWord.
+Require Import coqutil.Word.Naive.
+
+(* Local because it automatically adds arbitrary word size instance. *)
+#[local] Hint Extern 0 (word.word ?width) => exact (Naive.word width) : typeclass_instances.
+
+#[export] Instance word_ok {width} {BW : Bitwidth.Bitwidth width}: word.ok (Naive.word width).
```
lukaszobernig, line 10: OK I see, now we destruct here instead of explicitly re-exporting a corresponding `word*_ok`.

lukaszobernig (approved):
I like this approach of having to be very explicit about the used word size.

bedrock2/src/bedrock2/BasicCSemantics.v
miriampolzer: Ah, your comment reminded me that I had wanted to switch the Qed here for a Defined. Then you can actually show that the word_ok instances are equal when you instantiate a Lemma that was used with this instance in a context with a concrete bitwidth.

bedrock2/src/bedrock2/BasicCSemantics.v
```diff
@@ -0,0 +1,85 @@
...
+  assumption.
+Qed.
+#[export] Instance ext_spec_ok width {BW : Bitwidth.Bitwidth width}:
+    Semantics.ext_spec.ok (fun _ _ _ _ _ => False).
+Proof.
+  constructor; intros; try contradiction.
+  apply weaken_ext_spec.
+Qed.
+
+Section TypeclassTests.
```
samuelgruetter, line 45: Ah that's great to add some `TypeclassTests`, I should have done so in the first place!

PR #552, opened 2026-08-18
solve [] instead of eabstract for variable side conditions

### LINKED ISSUE #507: investigate straightline slowdowns for resolving Loops.enforce
I've seen at least two instances where straightline becomes really slow resolving Loops.enforce. I am guessing this is because there is a non-[]-restricted cbv on values, but have not verified that is the cause.
Latest instance can be reproduced here:
https://github.com/mit-plv/bedrock2/commit/a02822e1413dd5686b3d55c4f22b392d358fd164

miriampolzer:
Closes #507

### DIFF
--- a/bedrock2/src/bedrock2/ProgramLogic.v
+++ b/bedrock2/src/bedrock2/ProgramLogic.v
@@ -342,7 +342,7 @@ Ltac straightline :=
     change G;
     unshelve (idtac; repeat match goal with
                      | |- Markers.split (?P /\ Markers.right ?Q) =>
-                       split; [eabstract (repeat straightline) | change Q]
+                       split; [solve [repeat straightline] | change Q]
                      | |- exists _, _ => letexists
                      end); []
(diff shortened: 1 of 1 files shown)


miriampolzer:
@samuelgruetter do you have time to review this?

samuelgruetter:
I believe that `eabstract` was an attempt to speed up some slow `Qed`s. If I understand correctly, the change you're proposing speeds up some `straightline` calls. So I believe this is a performance tradeoff. I am fine with both versions, and would prefer the one that is fastest for those proofs that the most active users care most about, and I believe that's currently you, so if you verified that all proofs that you care about still build fast enough or faster, then this is good to merge for me!

miriampolzer:
Thanks! I haven't compared the numbers, but building all of fiat crypto with this change seemed to have equivalent runtime (speedups are mostly locally for the proofs we're currently building). I'll merge this, and also observe the CI speed after to ensure there's no noticable regression.

PR #561, opened 2026-09-12
Adapt to coqutil Word.Naive on stdlib bits; drop Rocq 9.0 from CI

JasonGross:
Adapts to coqutil `Word.Naive` being represented as stdlib `bits` (mit-plv/coqutil#186): pins `deps/coqutil` to that branch, rewrites the two division-by-zero cases and three shift cases in `compiler/src/compiler/NaiveRiscvWordProperties.v` (the only file that unfolded the old record), and adjusts CI. Rocq 9.0 and 9.1 are dropped from CI: their `rocq/rocq-prover` docker images ship rocq-stdlib 9.0.0 (no `Zmod`), so CI tests 9.2 and dev; the opam bound stays `rocq-stdlib >= 9.1`, which does allow Rocq 9.1. Everything in `make all` (compiler, processor, end2end, LiveVerif) builds; no other source changes were needed.

Depends on mit-plv/coqutil#186; CI cannot fetch the pinned coqutil commit until that merges, so this stays a draft until then.

Written by Claude Fable 5.1.


JasonGross:
Stacked follow-up: #562 (Kami words on stdlib bits via mit-plv/kami#51; its last four commits are the Kami part).

---

compiler/src/compiler/NaiveRiscvWordProperties.v
```diff
@@ -85,23 +86,16 @@ Section NaiveRiscvWord.
...
-      change Naive.wrap with (word.of_Z (word := (Naive.word (2 ^ Z.of_nat log2width)))).
-      match goal with
-      | |- context [?x =? ?y] => destruct (Z.eqb_spec x y)
-      end.
-      + eapply word.of_Z_inj_mod. cbn.
-        rewrite <- Zminus_mod_idemp_l.
-        rewrite Z_mod_same_full.
-        reflexivity.
-      + cbn in *. exfalso. congruence.
+      (* destr already substituted z := word.of_Z 0; word.divu y 0 is Zmod.udiv y zero = opp one *)
```
andres-erbsen: comments that narrate immediately adjacent code are not useful, please remove

JasonGross: Removed, and swept the whole diff of this PR and of the two stacked ones (#562, #563) for comments that narrate adjacent code or describe the change; the substance is in the commit messages now.

---

andres-erbsen (approved):

PR #567, opened 2026-09-17
Adapt to bitSlice moving from coqutil to riscv-coq

JasonGross:
Independent of mit-plv/coqutil#189 and meant to merge before it; no submodule change (`RiscvWordProperties` imports `riscv.Utility.Utility`, which provides `bitSlice` with both the old and the new riscv-coq). The e1000 receive-buffer size reads its two register bits with the standard library's `Zmod.slice`. One commit on master aa9ea11f; built bedrock2, compiler, processor and LiveVerif examples against master's own dependencies (and also against riscv-coq master 58937c33).

Written by Claude Fable 5.1.


andres-erbsen:
make sure this is up to date and passes CI

PR #568, opened 2026-09-17
Use lia directly instead of the blia alias

JasonGross:
Independent of mit-plv/coqutil#190 and meant to merge before it; no submodule change. Every `blia` is `lia` (55 files), and the 31 files that took `Lia` only through `coqutil.Z.Lia` now import it directly. One commit on master aa9ea11f; full `make all` against master's own dependencies.

Written by Claude Fable 5.1.


andres-erbsen:
@JasonGross-fiat-crypto-bot rebase again

PR #572, opened 2026-09-17
Adapt to riscv-coq's load/store interface on stdlib bits (option A)

JasonGross:
**Stacked on mit-plv/riscv-coq#60** (pinned here at its tip; the interface change is not backwards compatible), so both merge together; before mit-plv/coqutil#188. Draft until riscv-coq#60 is merged.

Option A of the tuple-based `LittleEndian` removal (alternative to mit-plv/bedrock2#566): with riscv-coq's interface on `bits`, `GoFlatToRiscv`, `RunInstruction`, `FlatToRiscvCommon`, `MMIO`, `AssemblyVerif` and `KamiRiscvStep` work on words directly; `memory_mapped_ext_calls_riscv` passes `bits.of_Z (8*n) (le_combine v)` to the riscv post and `le_split n (Zmod.unsigned v)` to `write_step` (no existential, no cast, no tuple); the `MemoryMappedExtCalls` specs are on `list byte` as in option B. No `tuple.cast`, `HList.tuple` or explicit-instance `to_list_of_list` rewrite is left in compiler, processor or end2end. One commit on master aa9ea11f (17 files), coqutil at master; built compiler (with examples), processor and end2end.

Written by Claude Fable 5.1.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01YJbiRBkJtM7x6PvYV6hmEm

Issue #477, opened 2025-07-16
supporting $@ (map.of_list_word_at) for stackalloc/stackdealloc


andres-erbsen:
bedrock2 stackalloc semanticsuse [anybytes](https://github.com/mit-plv/bedrock2/blob/master/bedrock2/src/bedrock2/Memory.v#L35-L39) which is defined in terms of [`$@`](https://github.com/mit-plv/bedrock2/blob/master/bedrock2/src/bedrock2/WeakestPrecondition.v#L84-L92) but the [ProgramLogic Ltac](https://github.com/mit-plv/bedrock2/blob/master/bedrock2/src/bedrock2/ProgramLogic.v#L171-L190) for discharging these goals still converts back to `array ptsto` for compatibility with existing proofs. For new proofs, we'd like to have an option where straightline on stackalloc creates `$@` hypotheses and expects the same for deallocation. Supporting this would probably come down to adding an alternative match-goal-with case in dealloc and probably having an alternative straightine_stackalloc that we can use to override the existing one using Ltac redefinitions (`::=`).

@lukaszobernig or @miriampolzer, this might be a good task to get more experience working with guts of bedrock2. I can also do it myself, sometime.

andres-erbsen:
@JasonGross-fiat-crypto-bot wdyt?

PR #573, opened 2026-09-18
Program logic: keep stack frames as maps on request (#477)

JasonGross:
Prototype for #477: on request, the program logic keeps a stack frame as the map `bs $@ a` from allocation to deallocation instead of converting it to `array ptsto (bits.of_Z width 1) a bs` and back.

Since #572 a scalar is `le_split n v $@ a` and `anybytes` is the existence of a `bs` with `bs $@ a = m`, so the map form composes with the scalar predicates by list rewriting alone; the new lemmas `scalar{,16,32}_of_list_word_at` and `truncated_{scalar,word}_of_list_word_at` are those bridges.

A proof opts in with a marker instance, `#[local] Instance : Memory.stackalloc_as_map := {}`, in its file or section. With it, `straightline_stackalloc` keeps the frame as `bs $@ a` (with `length_stack` and the `2 ^ width` bound in the context) and `straightline_stackdealloc` looks for the frame in that spelling first, seeing through the scalar predicates, and falls back to the array form. Without the instance both tactics are the previous ones, so no existing proof changes. The three program logics (plain, Metric, Leakage) carry the same tactics. `bedrock2Examples/stackalloc_of_list_word.v` is a small end-to-end example: stackalloc, store, load, dealloc, with one `seprewrite` to turn the frame into a `scalar32`.

Built: bedrock2/ (noex, all examples, special tests). Open points: the name and home of the marker class, and whether the old array route should stay the default.

Written by Claude Fable 5.1.


andres-erbsen:
@JasonGross-fiat-crypto-bot Search fiat-crypto and bedrock2 for workarounds to the lack of the feature added in this PR and make a stack of PRs to replace these workarounds with the feature here.

